PR #7, opened 2016-08-25
Fix #3 by using getLibraryFromClassname.

### LINKED ISSUE #3: Exception with Application named "wm.qx.mutablelist"
Hi John,

i try to compile an application named "wm.qx.mutablelist" which is a contrib and a demo at once,
it fails to compile that because the result of https://github.com/johnspackman/qxcompiler/blob/master/lib/qxcompiler/targets/Target.js#L228 is "wm" and not "wm.qx.mutablelist".

Do you have an idea how we could solve that?

pcdummy:

### DIFF
--- a/lib/qxcompiler/Analyser.js
+++ b/lib/qxcompiler/Analyser.js
@@ -182,10 +182,10 @@ module.exports = qx.Class.define("qxcompiler.Analyser", {
 
             // Find all classes
             function (cb) {
-              async.each(t.__libraries, 
+              async.each(t.__libraries,
                   function(library, cb) {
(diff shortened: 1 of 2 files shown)


johnspackman:
I dont think that this right - the namespace is not the classname, or even the class package.  EG you could have a namespace of "abc" and an application class of "abc.def.ghi.Application".  The namespace is defined in the Manifest.json, so we have to search for matches until we narrow down the library.

(PS Ive got back late after a couple of days of near-zero signal so just catching up but spending time on qx this holiday weekend)

pcdummy:
ok

PR #8, opened 2016-08-26
Allow assets with libraries with dotted names.

pcdummy:
Like: qxc.promise - source/resource/qxc.promise.


johnspackman:
The issue is that although “.” is legitimate in a namespace, it makes it impossible to determine the library without searching all libraries; qxcompiler is wrong because it is assuming that the first “.” separates the namespace, but I think the PR is being too aggressive.  For example in the class “abc.def.ghi.MyClass” or the asset “abc/def/ghi/image.png”, what’s the namespace?  That depends on Manifest.json, and without checking every library to find out, it could be “abc”, “abc.def”, or “abc.def.ghi”.

pcdummy:
We should add the namespace or the Library to every asset, then we can fix this, right?

johnspackman:
fixed in latest release

PR #9, opened 2016-08-27
Add option to copy resources in source builds.

pcdummy:
This makes serving the content with grunt-contrib-connect easier.


pcdummy:
We need a Hybrid version of this too, will make a general function for both.

PR #15, opened 2016-09-04
Remove trailing whitespaces.

pcdummy:


johnspackman:
If it has to search for the qooxdoo library path, it means that the path has been specified incorrectly; qxcompiler expects the path to be to the library (ie the folder where Manifest.json is) and not to the root of the repository - this is consistent with all other library paths, and therefore I'd say the current implementation is correct.

Im not sure if the problem you're trying to solve is that qxcompiler needs to use resources which are outside of the Qooxdoo library and is making assumptions about their location - they only exist if you have a complete copy of the repo.  That hasn't been a problem lately, but could be an issue when we make the npm `qooxdoo` module just the library.  As those resources are qxcompiler/generator specific, they could be stored within qxcompiler (as a copy from qooxdoo repo), and when generate.py is removed completely (eg in Qooxdoo v7) they would only exist in qxcompiler.

One other thing is that when you submit commits, please can you not include a lot of whitespace changes because it makes it difficult to see what's changed - I there's even one files that is only whitespace changes included in this patch

pcdummy:
Hey John,

I updated the commit on the same time you wrote the answer :)

Why i/we need this:
I've published [@pcdummy/qx](https://www.npmjs.com/package/@pcdummy/qx) which is the contents of [your qxcompiler-v2 branch + additions](https://github.com/pcdummy/qooxdoo/tree/master/framework), on change i made is that i copied parts of /tool/ to the framework folder as i publish just the framework folder and nothing else.

I see adding those resources to QxCompiler would be nice.

pcdummy:
Thought about adding resources to QxCompiler, not sure its a good idea with CLDR stuff also the bootloader, shouldn't the framework be able to change it without having to change QxCompiler?

johnspackman:
The loader and CLDR files are pretty closely tied to both qxcompiler *and* qooxdoo, I think it's 50/50 as to which part takes ownership of the files.  CLDR is compiled from public sources so it's not likely to be changed independently of qxcompiler, and as the loader.tmpl.js is directly used and processed by qxcompiler you might find that it should have its own copy anyway (there are already qxcompiler-specific mods in it)

pcdummy:
I'll remove the path changing patch here, can you then please merge the whitespace fix?

Next i'll come up with a patch that will include these files into QxCompiler, fine?

johnspackman:
Is the whitespace to do with eslint? (in which case can we turn it off?)

I'll merge it but IMHO as a general case you probably need to be able configure your editor to not modify whitespace - I have not deliberately added that whitespace at EOL which suggests that more will be created in the future.

pcdummy:
No, I configured to remove trailing Whitespaces. This is what we all have at my work, not sure i can make mine to ignore this behaviour on a specific project.

PR #33, opened 2017-09-29
Fixed creation/update of po files

mbgonicus:
I tried to fix that the compiler does not update the po files. This fix works, but I'm absolutely sure that this is pretty bad code... It probably does things at the wrong place, misuses idioms, lacks error handling and perhaps just fixes the symptoms - not the cause.

But hopefully it gives you an idea for what is really wrong or you can sanitize the code.

These are the key aspects that (I think) I have discovered:

1. `qxcompiler.Analyser.updateTranslations` seems to never be called, so a call is added to `qxcompiler.targets.Target`.
2. `qxcompiler.Translation#read` returns a promise, but it was expected to take and invoke a callback function.
3. In `updateTranslations` it is iterated over `dbClassInfo.messageIds`, but `dbClassInfo.translations` seems to be correct.
4. Also in `updateTranslations` `entry.comments` is empty. It now gets initialized with an empty object.
5. Again in `updateTranslations`, the class info is retrieved via `t.getClassInfo`, but that does not give a return value, but rather gives the result via a callback. Hence `translation.write` will not write the new entries, as they are not created yet. I'm using `Promise.all` to wait for the entries to be created; there might be a better way to do it.

I hope that his makes at least a litte bit of sense. I'm not really familiar with the code and got a little overwhelmed by it... ;)

### DIFF
--- a/lib/qxcompiler/Analyser.js
+++ b/lib/qxcompiler/Analyser.js
@@ -878,7 +878,7 @@ module.exports = qx.Class.define("qxcompiler.Analyser", {
 
     /**
      * Updates all translations to include all msgids found in code
-     * @param the library to update
+     * @param library the library to update
      * @param locales
      * @param cb
      */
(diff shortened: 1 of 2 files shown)


mbgonicus:
@johnspackman Did you find some time to look over it?

johnspackman:
@mbgonicus no not yet, im hoping to get to it and some other localisation stuff "real soon now" - sorry for the delay, but it looks like I have some business-related reasons to get back to translations and localisation

johnspackman:
@mbgonicus i've merged your changes manually, because I couldn't merge direct in github without causing a break upstream due to some of the class names had changed since you wrote your PR (so my fault for being so slow to get to it).  But apart from a minor class name change, everything's gone across as is.  thanks for the PR! 👍

PR #46, opened 2017-11-28
fix bug in cldr calculation

johnspackman:
Fixes a bug in CLDR; locales inherit from explicit parentLocale *and* implicit from their prefix (eg `en_gb` inherits from `en_001` and `en`)


johnspackman:
this isn't right and doesn't have unit tests, closing (to be replaced later)

PR #52, opened 2017-12-11
Fix async & await not being transpiled

johnspackman:
async and await were not being transpiled and would be simply copied into the output, this fixes that problem


johnspackman:
darn, managed to include a load of other commits in there, withdrawing this and I'll fix it separately

PR #53, opened 2017-12-11
fixes an issue where the detail from uglify error messages are lost, …

johnspackman:
…making it impossible to track down the underlying cause of a problem.


cboulanger:
Merged, thank you, @johnspackman

Issue #51, opened 2017-12-04
Build target seems to be broken


cajus:
If I run
```
$ qx compile --target=build
SyntaxError: Unexpected token keyword «function», expected punc «,»ogo
    at JS_Parse_Error.get (eval at <anonymous> (/home/cajus/Projekte/cajus/new/qooxdoo-compiler/node_modules/uglify-js/tools/node.js:21:1), <anonymous>:79:23)
    at qxcli.commands.Compile.process.catch (/home/cajus/Projekte/cajus/new/qooxdoo-cli/lib/qxcli/commands/Compile.js:610:31)
    at <anonymous>
```
while the *source* target works fine. *ogo* looks like a cropped version of the applications name "nogo".

johnspackman:
This is because you've checked out the latest github, but there are new modules which npm needs to download.  You need to `cd` to `qooxdoo-cli` and `npm i`; if you also have a checkout of `qooxdoo-compiler` you then also need to `npm link qxcompiler`

johnspackman:
It looks like it's the same app that was in the tarball you sent me a short while ago?  I've just tried with that and can't reproduce it; please can you confirm that both copies of the repo are up to date and using the `npm link` to get compiler hooked into cli, then delete `db.json`, `resource-db.json`, and `build-output/` and try again?  If it still fails, please can you send me an updated tarball and I'll try again?

cajus:
@johnspackman that's the same application, which is basically unchanged. I don't use a globally linked qx-cli/compiler. There's just qx-cli reference  in the *package.json*. Removing the *db.json files does not seem to help.

The playground itself doesn't work completely in the moment, because I'm playing around with *feathersjs* on the backend side. But while it's just a "build" issue, that doesn't matter, I guess.

Sending you the details via *gitter*.

Issue #44, opened 2017-11-20
Async and await are not transpiled!


hkollmann:
example:
original code:

    // Add an event listener
      button1.addListener("execute", e => {
         this.__testAsync();
      });
    },
    __testAsync: async function () {
       await dialog.Dialog.alert( "Hello World!" + "\n" + QueryString?"QueryString loaded":"QueryString not loaded").promise();
    },   
    __testAsync: function () {
       dialog.Dialog.alert( "Waited for await!");
    }

transpiled code:

        button1.addListener("execute", function (e) {
          _this.__testAsync();
        });
      },
      __testAsync: async function __testAsync() {
        await dialog.Dialog.alert("Hello World!" + "\n" + QueryString ? "QueryString loaded" : "QueryString not loaded").promise();
      }
    }, "__testAsync", function __testAsync() {
      dialog.Dialog.alert("Waited for await!");
    })

PR #54, opened 2017-12-11
Fixes a bug where async/await was not transpiled

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/44
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/51


cboulanger:
Merging, thanks @johnspackman - great to see we're using babel-preset-env now...

PR #55, opened 2017-12-11
fix bug in cldr calculation

johnspackman:
fixes bugs in Locale CLDR output (locale data can be inherited)


cboulanger:
Merging, thank you @johnspackman

Issue #56, opened 2017-12-14
Build target doesn't load


cajus:
That's in the browser console when the debugger stops the execution:

```
ReferenceError: regeneratorRuntime is not defined 
at http://localhost:3030/build/nogo/part-0.js:1:797181
at http://localhost:3030/build/nogo/part-0.js:1:797754
at http://localhost:3030/build/nogo/part-0.js:1:800162
```

PR #58, opened 2017-12-15
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/56, caused by missing babel polyfill

johnspackman:


cboulanger:
Why does the test fail with `rake failed`? Doesn't seem anything to do with our code...

hkollmann:
I think because travis is not configured. And then it uses the default configuration. Can we ignore this?

hkollmann:
Yes we can. Merge is allowed.

cboulanger:
But shouldn't we fix this first? It's probably not difficult since the tests used to pass, didn't they @johnspackman ?

cajus:
I've already prepared a PR that includes travis and npm support for this repo. Looks like @oetiker didn't tick the checkbox for "only run if .travis.yml is present".

cajus:
Currently the PR fails in travis, because 'npmt test' fails. See https://travis-ci.org/cajus/qooxdoo-compiler

cajus:
Just fixed the travis settings.

hkollmann:
Thanks  @johnspackman!

PR #59, opened 2017-12-15
Initially added travis file

cajus:
Note that the email might not be correct. And the API token definitely is not correct and has to be updated by @johnspackman after it's merged.

Currently the `npm test` run fails. Not sure if it is suppsed to fail and just needs to be disabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: node_js
+node_js:
+- '8'
+cache:
+  directories:
+  - node_modules
+before_install:
+- git clone https://github.com/qooxdoo/qooxdoo.git ../qooxdoo
+deploy:
+  provider: npm
+  email: johnspackman@zenesis.com
+  api_key:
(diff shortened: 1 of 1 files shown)


cajus:
That's the task of the common "new version" preparal. I'd guess thats the responsibility of the maintainer who's also doing the tagging in git. If you do that automatically, the git repo with your tag and the contents of the package would not be in sync.

cajus:
Also it would require that you update the tag from within travis (pass credentials, etc.). Updating the tag would trigger travis again.

cboulanger:
I think that's why we have semver. Unless you specifically update the major (breaking) version, you can do automated (patch) updates and everything will stay in sync - since semver guarantees that any "minor" or "patch" update is backwards compatible. Of course, this has to be actually the case and needs to be enforced.

cajus:
@cboulanger sorry, no idea what you want to say ;-) Fact is that a "git tag" triggers all that. And this requires that everything already is in shape for deploying.

hkollmann:
OK, understand after reading the docs.
publish is only started if a tag is set

cajus:
Good point. I guess using https://github.com/qooxdoo/qooxdoo/archive/master.zip would be more appropriate in this case anyway anyway.

hkollmann:
Or a npm install after we released the sdk

cajus:
Which does not work now. And the tests require to have qx in `../qooxdoo`.

hkollmann:
Your right. How is master.zip generated?

cajus:
Github generates it. I've updated the PR to use it (a tar.gz in this case) directly.

cboulanger:
Could this be the solution to our problem?: https://www.npmjs.com/package/ci-publish

cboulanger:
Also interesting: https://oncletom.io/2016/npm-release/

johnspackman:
3 approvals.  I'm merging this despite the travis failure because there is an issue in the code that has nothing to do with this PR.  thank you @cajus !

PR #61, opened 2017-12-15
Add path to uri mapping

johnspackman:
This allows the paths to resources, transpiled code, etc to be arbitrarily mapped to URIs to match the web server configuration.  See upcoming CLI branch for examples and documentation.

EDIT: This is based on https://github.com/qooxdoo/qooxdoo-compiler/pull/58


hkollmann:
Thanks @johnspackman!

Issue #49, opened 2017-11-30
Mixin with environment dependent definition is not parsed


mbgonicus:
We have a mixin which is defined similar to this:

```javascript
qx.Mixin.define("foo.MBar", qx.core.Environment.select("qx.baz", {
  "true" : { /* ... */ },
  "false" : { /* ... */ }
}));
```

It used to work with the old python generator, but with the new compiler the compilation is aborted with the message:
```
foo.MBar: Missing class definition - no call to qx.Class.define (or qx.Mixin.define etc)
ERROR: Cannot write application 'foo' because it has fatal errors
```

I traced it down to [ClassFile.js#L828](https://github.com/qooxdoo/qooxdoo-compiler/blob/bd3b4a80c720a16c1334dfdd5dcdfd773b0be234/lib/qxcompiler/ClassFile.js#L828):

```javascript
if (classDef.type != "ObjectExpression") {
  return;
}
```

As `classDef.type` is a `CallExpression`. 

A workaround would be to move this check below [ClassFile.js#L836](https://github.com/qooxdoo/qooxdoo-compiler/blob/bd3b4a80c720a16c1334dfdd5dcdfd773b0be234/lib/qxcompiler/ClassFile.js#L836):
```javascript
t.__foundClassDef = true;
```
This prevents the error and at least the compilation is not aborted. But of course the definition is not parsed and therefore not considered for dependencies, translations, etc.

I don't have enough understanding of the code base to fix this. But even without trying I'm sure it also effects Class and Interface definitions as well.

mbgonicus:
It seems that the functionality is basicly there: [ClassFile.js#L938](https://github.com/qooxdoo/qooxdoo-compiler/blob/01f623113cba0bf8daf4e7b417e0e7568ff27819/lib/qxcompiler/ClassFile.js#L938)

But there is a todo comment. What exactly is there todo?

johnspackman:
it's for code elimination - so if the call to `qx.core.Environment.select` can be determined at compile time, the code is modified to only have the code that will apply.  I've done it for `.get` but it looks like I forgot to go back and implement `.select`

mbgonicus:
I don't really get it - the code for `.get` and `.select` seems to be identical. Does it work for `.get`?

johnspackman:
You're right - somehow the code for code elimination has been, erm, eliminated!  I'll have to dig it out and reinstate it.

PR #67, opened 2018-01-08
Implemented substitution of class definition if in environment select check

mbgonicus:
PR for issue #49 

As always, I'm not really sure if this is a good way to do it. The compile is still somewhat mysterious to me... But it works in an application that we use. 

Feel free to edit.

### DIFF
--- a/lib/qx/tool/compiler/Analyser.js
+++ b/lib/qx/tool/compiler/Analyser.js
@@ -92,6 +92,12 @@ module.exports = qx.Class.define("qx.tool.compiler.Analyser", {
       init: false,
       nullable: false,
       check: "Boolean"
+    },
+
+    /** Environment during compile time */
+    environment: {
+      init: null,
+      check: "Map"
     }
   },
 
--- a/lib/qx/tool/compiler/ClassFile.js
+++ b/lib/qx/tool/compiler/ClassFile.js
@@ -552,7 +552,39 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
           }
         }
       };
-      
+
+      const CODE_ELIMINATION_VISITOR = {
+        CallExpression(path) {
+          const name = collapseMemberExpression(path.node.callee);
+
+          if (name === "qx.core.Environment.select" || name === "qx.core.Environment.get") {
+            const arg = path.node.arguments[0];
+            const env = Object.assign({}, qx.tool.compiler.ClassFile.ENVIRONMENT_CONSTANTS,
+              t.__analyser.getEnvironment());
+            const envValue = env[arg.value];
+
+            if (envValue !== undefined) {
+              if (name === "qx.core.Environment.get") {
+                path.skip();
+                path.replaceWithSourceString(envValue.toString());
+              } else if (name === "qx.core.Environment.select") {
+                const subPath = path.get('arguments.1');
+                let option = subPath.node.properties.find(prop => prop.key.value === envValue.toString());
+
+                if (!option) {  // try to find default value
+                  option = subPath.node.properties.find(prop => prop.key.value === "default");
+                }
+
+                if (option) {
+                  path.skip();
+                  path.replaceWith(option.value);
+                }
+              }
+            }
+          }
+        }
+      };
+
       function collectJson(node) {
         var result;
         if (node.type == "ObjectExpression") {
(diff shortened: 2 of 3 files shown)


johnspackman:
this looks good to me but i want to try merging it with my code and trying ti out first - i have a small backlog of qxcompiler stuff that i'm working through, this is one.  sorry it's taking a while!

mbgonicus:
@johnspackman I'm trying to revise atm. Is it guaranteed that the values checked by qx.core.Environment.get/select are set at compile time? Or in other words: may I always eliminate the code or are there cases where the select/get things must retain for runtime?

mbgonicus:
>The solution is to move your code elimination to ClassFile.js#L925-L952. As the parser iterates over each node in the AST tree, it eliminates/reduces the code wrapped in qx.core.Environment calls (if possible), and by the time it gets to L869 the reduction has already happened.

>It would be preferable if the code at L869 made a check that the type of the defineArg is an ObjectExpression, and if not it should raise an error.

@johnspackman I moved the elimination over to ~ L 925 as you mentioned. It now works for calls of qx.core.Environment.select, like in [qx.Class#L759](https://github.com/qooxdoo/qooxdoo/blob/382e296ec029845d7903b361a5ccd8943a3d55c9/framework/source/class/qx/Class.js#L759), but not for class definitions.

When I'm reaching [ClassFile.js#L869](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/lib/qx/tool/compiler/ClassFile.js#L869) (CallExpression visitor), the code elimination has not occured yet. It will visit the `qx.Class.define` before it visits the `qx.core.Environment.select` piece. Therefore, it still produces  the error

```
Missing class definition - no call to qx.Class.define (or qx.Mixin.define etc)
```

still occurs. 

How can I make the code elimination before the `qx.Class.define` is being handled?

johnspackman:
> What do you mean by defineArg?

That was the name of the variable you used in your original PR

> How can I make the code elimination before the qx.Class.define is being handled?

Babel plugins are written as a series of Visitors (eg [VISITOR at ClassFile.js#L765](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/lib/qx/tool/compiler/ClassFile.js#L765) and [CLASS_DEF_VISITOR on ClassFile#L610](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/lib/qx/tool/compiler/ClassFile.js#L610) and [COLLECT_CLASS_NAMES_VISITOR  at ClassFile.js#L544](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/lib/qx/tool/compiler/ClassFile.js#L544)) - the visitor is executed on every node of the AST tree recursively.

Generally speaking the principal is that context is not allowed in a visitor (because it often breaks), but we can process visitors in order and sometimes switch from one visitor to another; VISITOR is the main visitor, and it delegates to the other two as needed (note that when handing off to another visitor, `path.skip()` is used to stop the current visitor from continuing to process the current node and it's children).

So the code elimination needs to go in an outer visitor, so that when the inner visitor sees the code the transformation has already happened.  That might mean a little refactoring of the code that's there :)

mbgonicus:
Ok, so if i understand it correctly, there must be a separate "code reduction visitor"? That will be quite a bit of work (at least for me). Need to find some time to do that...

johnspackman:
> so if i understand it correctly, there must be a separate "code reduction visitor"?

Yes, at this stage I think that's right.  It's an essential feature though, so one of us will get there first :)  In the mean time, your original patch will work as a work around for you if you're happy to keep following master

mbgonicus:
All right, I think I got it working now. I push-forced because rebase because master has changed a lot. Might investigate in other code reductions like qx.core.Environment.filter, but for select and get it will do.

Hope that is how you intended?

johnspackman:
@mbgonicus thank you for your work on this PR - I've made a couple of minor adjustments and added some unit tests at https://github.com/johnspackman/qooxdoo-compiler/tree/fix-env-subs, please can you take a look and either consider integrating them into your PR or I can do a separate PR (with full credit obviously! 👍 )

mbgonicus:
@johnspackman Thank you for investigating! I'm fine with your changes. It would be the easiest way if you just make a PR based on your branch.

One little issue: isn't it necessary to add an [environment property to Analyser.js](https://github.com/qooxdoo/qooxdoo-compiler/pull/67/files#diff-82960fce90273847b0f7130d054ee37cR97)? Otherwise it wouldn't be possible to [set the environment in AppMaker.js](https://github.com/johnspackman/qooxdoo-compiler/commit/59d92c58d7c74633eacb376539d1fc0ea9ccf44e#diff-0e2f993d09366a511d2a184f83aa21ebR85)?

johnspackman:
You're absolutely right, it is!  Something went wrong with my commit and I lost some of the changes, including the environment settings, my branch has just updated with the correction.  

I'll close this and make a new PR

PR #75, opened 2018-01-16
Improve initial (template) index.html

cajus:
Tweak parts of the *mobile* index.html and copy it over to *desktop*. There is no reason, why these two should be treated differently. If it is omitted, the use experience on tablets (which are valid targets for desktop applications) is really poor. This way, you can link a qx desktop based application to your tablets home screen, too.

Add a noJS fallback, to inform users that they should enable JS instead of letting them alone with a white, blank page.


johnspackman:
3 approvals, merging, thank you @cajus !

PR #90, opened 2018-01-17
Fixed writing of translation entries

mbgonicus:
As you can see in [Target.js L634](https://github.com/mbgonicus/qooxdoo-compiler/blob/8831b76505d56ec1c850ad0015eefbf22a434dfe/lib/qx/tool/compiler/targets/Target.js#L634), the function `writeEntry` has the `localeId` as its first parameter:

```javascript
function writeEntry(localeId, entry) {
```

The call of `writeEntry` in the corrected lines did not state the `localeId` but just the entry. Therefore, the translations are never written to `compileInfo` and never arrive in the compiled application.


johnspackman:
3 approvals, merging, thank you @mbgonicus !

PR #91, opened 2018-01-17
first release of `qx serve` based on metalsmith and new qooxdoo theme

johnspackman:


cboulanger:
Can you quickly outline how it works? :-)

johnspackman:
oops, missing documentation!!  I'll do that in a mo, but in the mean time: running `qx serve` compiles the app and starts a web server on `http://localhost:8080`.  The compilation is actually a continuous compile, so it's the equivalent of `qx compile --watch` plus a web server; all the usual parameters to `qx compile` apply (except `--watch`, which is always implied).

You can change the port it listens on by using `--listenPort 8082` or adding `serve: { listenPort: 8082 }` to compile.json (the command line takes precedence over the `compile.json`

johnspackman:
also it shows up using @cajus 's new theme :) 
screen shot 2018-01-17 at 13 42 20

lib/qx/tool/cli/commands/MConfig.js
```diff
@@ -20,139 +20,139 @@
...
+      }
+      return config;
+    },
+
+    /* 
+     * Merges the argv into the config data 
+     * 
+     */
+    mergeArguments: function(parsedArgs, config, contribConfig) {
+      if (parsedArgs.config) {
```
hkollmann: mergeArguments:
Should'nt this not be protected? Or is it used from outside?

lib/qx/tool/cli/commands/Serve.js
```diff
@@ -0,0 +1,224 @@
...
+        config.serve = {};
+      var serve = config.serve;
+      serve.listenPort = serve.listenPort||this.argv.listenPort;
+      return config;
+    }
+  },
+  
+  statics: {
+    
+    getYargsCommand: function() {
```
hkollmann: I all the other files this is in the beginning of the file...

hkollmann:
aproved and merged.
Thanks @johnspackman

PR #97, opened 2018-01-19
fixes for node support, unifies rhino and node

johnspackman:


hkollmann:
Thanks, @johnspackman !

Issue #74, opened 2018-01-16
Add (npm) scripts to a qooxdoo app


cboulanger:
@derrell and I have talked about how to most efficiently add external non-qooxdoo libraries to qooxdoo applications. Here is what we came up with. We add two commands:

1) `qx add script path/to/script.js [--path=<resource path, defaults to source/resource/js>]`: 
adds a script to the application setup (copying it into resources, adapting the `externalResources` in Manifest.json)
2) `qx add npm-package <npm package name> [—entry=<entry script, defaults to browser.js>], which does the following:
- npm install <package name>
- look for the entry script (default:browser.js)
- calls webpack with the option to create a bundle named as the npm package, which will be exposed as a global variable with the name of the npm package (dashes converted to underscores)
- calls `qx add <bundle file>`

for example, if you need a JsonRpc client (that was the initial question), you could do `qx add npm-package raptor-client` to install https://github.com/LinusU/raptor-client and then, in your qooxdoo code, you can do `let raptor = window.raptor_client; let client = new raptor(url)...`

derrell:
The idea here is that any browserify-able or webpack-able NPM package can be trivially added to a desktop qx app, and used. That would *greatly* expand the available capabilities for a qx app.

hkollmann:
Great idea. Would make adding npm packages much more easy
Needs an option to update the Version in some kind

PR #105, opened 2018-01-22
Implementation of "qx add script"

cboulanger:
See #74 
```text
qx add script
qx add script <scriptpath> [options]

Options:
  --resourcedir, -d  The subdirectory of the resource folder in which to place
                     the file                                    [default: "js"]
  --rename, -r       Rename the file to the given name
  --undo, -z         Removes the file that would normally be added with the
                     given arguments
```


hkollmann:
2 approvals, merged.
Thanks @cboulanger!

PR #106, opened 2018-01-23
fixes two bugs in `qx serve`, one that was preventing continuous comp…

johnspackman:
…ile, and the other with output paths


hkollmann:
Thanks @johnspackman !

Issue #107, opened 2018-01-23
Rework contrib demo app architecture


cboulanger:
In the old contrib system, demos were placed in the `demo/default` subfolder as a separate application with a separate Manifest. The new `qx contrib` system has continued to support that layout. However this has a couple of disadvantages and quirks:

- there are two `Manifest.json` files which have to be kept in sync, with no benefit;
- the library doesn't have a `compile.json` and the `qx contrib` commands have to look in the demo subfolder to get information on the qooxdoo version;
- there has to be an additional file `qooxdoo.json` to provide information on the place of the demo application;
- the additional cognitive overhead of having to deal with two libraries (the contrib library and the application library). 

This should be changed. Demo apps should be part of the contrib library namespace. The compiler can then pick up the application info and create a nice landing page with `qx serve`

cboulanger:
This also needs to be changed in the contrib library skeleton.

johnspackman:
the contrib's `Manifest.json` can "advertise" one or more applications, which the user can then choose to add into their compile.  This would mean that a contrib author could offer these kind of instructions to anyone wanting to try a debuggable version of their widget:

```
$ qx contrib add myusername/reponame
$ qx enable myusername/reponame/demoapp
$ qx serve
```

When the user visits http://localhost:8080 they would see the new demo app listed.

This is useful for more than just demo apps for contribs - for example, we want to move the API viewer into a separate repo, the API viewer *must* be compiled as part of the current build in order to view the classes being coded by the user, so this mechanism would make it zero-configuration to add API viewer too.

johnspackman:
IMHO the `qx enable` syntax is a bit long winded and forces contrib authors to maintain a unique application ID for each application class; using this as a unique ID means that we have to handle duplicate IDs (how many ids would be "demoapp" for example).

Instead, we should use the classname of the demo app - as the demo class is part of the contrib codebase it's immediately accessible and it's obvious what's being added:

```
$ qx contrib install cboulanger/qx-contrib-Dialog
$ qx enable dialog.demo.DemoApp
$ qx serve
```

Also, in the contrib Manifest.json, we could support a syntax like this:

```
{
  "provides": {
    "namespace": "dialog",
    "encoding": "utf-8",
    "class": "source/class",
    "applications": [ 
      {
        "class": "dialog.demo.DemoApp",
        "auto-enable": false
      }
    ]
```

hkollmann:
Done in actual version

PR #108, opened 2018-01-23
This moves the demo app in the contrib library skeleton in accordance…

cboulanger:
… with #107


hkollmann:
Thanks, @cboulanger!

PR #113, opened 2018-01-25
Add qx clean

cboulanger:
Adds the `qx clean` command which removes target output directories and the `*db.json` cache files.


hkollmann:
Thanks, @cboulanger!

PR #119, opened 2018-01-31
Add server template

hkollmann:


tool/cli/templates/skeleton/server/readme.tmpl.txt
```diff
@@ -0,0 +1,36 @@
+SERVER Skeleton - A qooxdoo Application Template
+================================================
+
+This skeleton application was created by running the 'create-application.py'
```
cboulanger: Needs to be updated to `qx create`

tool/cli/templates/skeleton/server/source/class/custom/test/DemoTest.tmpl.js
```diff
@@ -0,0 +1,68 @@
...
+   License:
+
+   Authors:
+
+************************************************************************ */
+
+/**
+ * This class demonstrates how to define unit tests for your application.
+ *
+ * Execute <code>generate.py test</code> to generate a testrunner application
```
cboulanger: Also here the references to the old generator should be updated.

tool/cli/templates/skeleton/server/source/translation/readme.txt
```diff
@@ -0,0 +1,3 @@
+This directory will contain translation (.po) files once you run the
```
cboulanger: Same here...

lib/qx/tool/compiler/app/loader-server.tmpl.js
```diff
@@ -155,15 +155,9 @@
       },
 
       signalStartup : function() {
-        qx.Bootstrap.executePendingDefers();
-        qx.$$loader.delayDefer = false;
```
johnspackman: I don't think that these can be removed, what's your rationale?

lib/qx/tool/compiler/app/loader-server.tmpl.js
```diff
@@ -155,15 +155,9 @@
...
         qx.$$loader.scriptLoaded = true;
-        if (window.qx && qx.event && qx.event.handler && qx.event.handler.Application) {
-          qx.event.handler.Application.onScriptLoaded();
-          qx.$$loader.applicationHandlerReady = true;
-        } else {
-          qx.$$loader.applicationHandlerReady = false;
-        }
+        qx.core.BaseInit.ready();
+        qx.$$loader.applicationHandlerReady = true;
```
johnspackman: `qx.event.handler.Application.onScriptLoaded();` is valid even for server apps because server apps derive from `qx.application.Basic`; however I think I recall that there needs to be a so fix that `qx.application.IApplication` has an `@require(qx.event.handler.Application)` that the first path of this "if" statement is activated properly.

If there isn't an application, then I can see that `qx.core.BaseInit.ready()` should be called, but that would be the `else` statement, not replacing the application startup

hkollmann: Even if you patch qx.application.IApplication with @require(qx.event.handler.Application)  constructor qx.event.handler.Application is not called so that  $$instance is undefined. So onScriptLoaded did not work.

I just copied from the loader of the current framework

lib/qx/tool/compiler/app/loader-server.tmpl.js
hkollmann: You are right - this solves the problem that qx.bom.client.Browser is'nt loaded. Thanks for the hint

lib/qx/tool/compiler/app/loader-server.tmpl.js
hkollmann: That are nearly the same results i debugged yesterday. qx.core.BaseInit.ready is just called as result of an windows Event. So in server enviroment it must be called direct.

hkollmann:
After adding runtime tests work is finished. Main idea was adding runtime test.

johnspackman:
2 approvals, merging, thank you @hkollmann !

johnspackman:
@hkollmann in https://github.com/qooxdoo/qooxdoo-compiler/commit/9a10e2a75fa55b5707242f9b7291bfafd8089331, why did you make the server applications start to respect source uris and resource uris?  

In both cases, they should be ignored in server apps because there is no such thing as a URI for alternative load paths

hkollmann:
I think you are right - copied this from the other loader

Issue #121, opened 2018-01-31
qx create --type server


hkollmann:
If you create an server application and compile it
`
qx create myapp --type server -I
qx compile
`
you get boot.js and index.html in myapp dir. In this case index.html is not needed and boot.js should be renamed to myapp.js so that it can be run in node with this name.

@johnspackman: If you have a hint where this is handeled i can  prepare a PR.

hkollmann:
closed with merge of   #126

Issue #64, opened 2018-01-01
index.html should be at top level, parallel to `resources` directory


derrell:
The relative path to the resources directory from the path where `index.html` was found should not be `../resources` as currently implemented; it should be `./resources`. The `index.html` files should therefore be placed at the top-level, not inside of the "application-name" directory where `part-*.js` are located.

derrell:
Hmmm... Then you can't have more than one target existing at one time... but still, the resources path should be relative to index.html. Maybe there should be a `build-output/resources` symbolic link to `../resources' ?

cajus:
I guess that is a duplicate of https://github.com/qooxdoo/qooxdoo-compiler/issues/57

derrell:
Yes, very similar to #57. My second comment may be more efficient -- create a symlink to the higher-level resources directory -- so that resources needn't exist in all target directories. Then again, I'm linux-centric, and symlinks are not commonly used on windows. I guess putting everything below the app directory may be more appropriate.

cajus:
From my point of view, it should output the same structure than before. @johnspackman said that he want's to make it configurable.

To be honest - being forced to use symlinks in the file system or aliases/rewrites in the web server configuration feels a bit odd. I'd expect the entry point - as you - to be at the top level. Except you've a multi application setup with shared resources - but that should not be the default.

Let's see what happens ;-)

johnspackman:
I've posted a comment on (57)[https://github.com/qooxdoo/qooxdoo-compiler/issues/57#issuecomment-355224979] but to address a couple of points raised here - the paths should all be relative to the application's `index.html`, if it's not then it's definitely a bug.

Also re: symlinks: this is possible on Windows but only when running as administrator, and there is no convenient `sudo` equivalent AFAIK.  I'm not sure it's necessary though, provided that it's OK to expose the entire target output directory

hkollmann:
Fixed with #127

Issue #57, opened 2017-12-14
Build file/directory layout needs extra HTTP server configuration


cajus:
If you generate your app with the proposed minimum configuration, it creates this directory structure:

```
├── app
├── resource
│   ├── leaflet
│   │   └── images
│   ├── nogo
│   │   ├── font
│   │   └── image
│   └── qx
│       ├── decoration
│       │   ├── Indigo
│       │   │   └── font
│       │   └── Simple
│       │       ├── arrows
│       │       ├── checkbox
│       │       ├── colorselector
│       │       ├── cursors
│       │       ├── menu
│       │       ├── splitpane
│       │       ├── table
│       │       ├── tabview
│       │       ├── tree
│       │       ├── treevirtual
│       │       └── window
│       ├── icon
│       │   └── Tango
│       │       └── 16
│       │           ├── actions
│       │           ├── apps
│       │           ├── mimetypes
│       │           └── places
│       └── static
```

This is something that cannot be served directly, because you normally don't want `http://foo.bar/app` as a base. The resulting structure should be self containing like it is in current qx applications:

```
└── app
    └── resource
        ├── leaflet
        │   └── images
        ├── nogo
        │   ├── font
        │   └── image
        └── qx
            ├── decoration
            │   ├── Indigo
            │   │   └── font
            │   └── Simple
            │       ├── arrows
            │       ├── checkbox
            │       ├── colorselector
            │       ├── cursors
            │       ├── menu
            │       ├── splitpane
            │       ├── table
            │       ├── tabview
            │       ├── tree
            │       ├── treevirtual
            │       └── window
            ├── icon
            │   └── Tango
            │       └── 16
            │           ├── actions
            │           ├── apps
            │           ├── mimetypes
            │           └── places
            └── static
```

johnspackman:
I'm thinking about how to handle this, but the root directory in the example is the target output directory, ie `source-output/` or `build-output/` and I've anticipated that it's the entire target output directory that would be published on the web server; eg you'd publish `http://foo.bar/build-output` as a base and then `http://foo.bar/build-output/app/index.html` would be the URL for the application.  The URIs used by the application are (or should be) all relative, so the `build-output` can be renamed either by modifying `compile.json` or by just renaming it as part of deployment (i.e. without a compile step).

[...]

PR #122, opened 2018-02-02
Fix index.html

hkollmann:
#121: In case of type node or rhino index.html is not generated and boot.js is named to application.js
#57, #64: Added a new parameter to compile.json, application: writeIndexHtmlToRoot. If set to true an <base href="appdir" /> is written to index.html. So the application can be used in the root of the webserver.

### DIFF
--- a/lib/qx/tool/compiler/app/Translation.js
+++ b/lib/qx/tool/compiler/app/Translation.js
@@ -242,6 +242,7 @@ module.exports = qx.Class.define("qx.tool.compiler.app.Translation", {
                 var value = m[2];
                 if (value.length >= 2 && value[0] == '"' && value[value.length - 1] == '"') {
                   value = value.substring(1, value.length - 1);
+                  value = value.replace(/\\t/g, "\t").replace(/\\r/g, "\r").replace(/\\n/g, "\n").replace(/\\"/g, '"');
                   set(key, value);
                 }
               });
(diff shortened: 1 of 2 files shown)


lib/qx/tool/cli/commands/Compile.js
```diff
@@ -326,6 +326,7 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
       if (t.argv["app-name"])
         appNames = t.argv["app-name"].split(',');
       
+      let hasIndexHtmlToRoot = false;
```
cajus: The default should match the *current* behaviour. Which would be `true`.

hkollmann: Default behaviour to generate.py or to new compiler? For new compiler default is false.

cajus: I'd expect the new way to behave more or less like the old way. Mapping to a multi-application layout by default is a special usecase IMHO. As written in the issue itself, I'd not expect to be forced to some subdirectories.

hkollmann: what's about ``` let hasIndexHtmlToRoot = (data.applications.length === 1);'''
Then we have root in a single application and will not break multi applications.

We will break all settings from people wich use new compiler.

cajus: Whatever retains the 'old' beaviour in some way is welcome. There has not been any multi application setup before - at least not without tweaking around in the *config.json* file.

So if it's possible to keep people with a standard single application setup from facing a 'strange' file layout (at least for the most common usecase), would be really nice.

Alternatively the default `qx create application` command should create a configuration which is configured for single applications properly - but I still think that the default value for such a setting should be the most common one. Developers who want to create a multi application layout can configure it.

hkollmann:
Just deleted my changes with git... Will reopen on monday

Issue #123, opened 2018-02-02
escaping po files


hkollmann:
Having a text like tr("Rechte für den Benutzer "%1"") leads to following msgid:
msgid "Rechte für den Benutzer "%1""

This should be

msgid "Rechte für den Benutzer \\"%1\\""

hkollmann:
Here are the escapes from polib.py:

```
def escape(st):
    """
    Escapes the characters ``\\\\``, ``\\t``, ``\\n``, ``\\r`` and ``"`` in
    the given string ``st`` and returns it.
    """
    #return st.replace('\\', r'\\')\
    return st\
             .replace('\t', r'\t')\
             .replace('\r', r'\r')\
             .replace('\n', r'\n')\
             .replace('\"', r'\"')

# }}}
# function unescape() {{{

def unescape(st):
    """
    Unescapes the characters ``\\\\``, ``\\t``, ``\\n``, ``\\r`` and ``"`` in
    the given string ``st`` and returns it.
    """
    def unescape_repl(m):
        m = m.group(1)
        if m == 'n':
            return '\n'
        if m == 't':
            return '\t'
        if m == 'r':
            return '\r'
        if m == '\\':
            return '\\'
        return m # handles escaped double quote
    #return re.sub(r'\\(\\|n|t|r|")', unescape_repl, st)
    return re.sub(r'\\(n|t|r|")', unescape_repl, st)

# }}}
```

hkollmann:
according to polib this item's must be escaped/unescaped:

            # Take care of keywords like
            # msgid, msgid_plural, msgctxt & msgstr.

hkollmann:
closed with merge of #125

PR #125, opened 2018-02-03
fix for #123: escaping po files

hkollmann:


johnspackman:
2 approvals, merging thank you @hkollmann

PR #126, opened 2018-02-04
fix for #121: if type is server app set name of boot.js to application.name and do not generate index.html

hkollmann:


johnspackman:
two approvals, merging thank you @hkollmann

PR #127, opened 2018-02-04
fix for #57, #64: write base tag to index.html

hkollmann:
Add a new parameter to compile.json, application: writeIndexHtmlToRoot. 
If set to true <base href="appdir">/ is written to index.html. 
So the application can be used in the root of the webserver.

### DIFF
--- a/docs/cli/compile-json.md
+++ b/docs/cli/compile-json.md
@@ -67,6 +67,7 @@ The `applications` key is an array of objects, and each object can contain:
 application.
 `loaderTemplate` - (**optional**, **advanced**) this is the boot loader template file, usually determined automatically from the application `type` 
 `minify` - (**optional**) determines the minification to be used for this application, if the target supports it; overrides other settings.  Can be `off`, `minify`, `mangle` or `beautify`; takes precedence over the target's `minify` setting.
+- `writeIndexHtmlToRoot` - (**optional** ) if true the index.html file will be written to the target output directory. Allowed only once if you have more then one application. The index.html get's an <base href="appname" /> if set to true.
 
 A complete example is:
 ```
--- a/lib/qx/tool/cli/commands/Compile.js
+++ b/lib/qx/tool/cli/commands/Compile.js
@@ -326,6 +326,7 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
       if (t.argv["app-name"])
         appNames = t.argv["app-name"].split(',');
       
+      let hasFoundIndexHtmlToRoot = false;  
       data.applications.forEach(function(appData, appIndex) {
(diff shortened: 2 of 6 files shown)


hkollmann:
Do you have an better name? 
I could make it default for single application configurations

johnspackman:
Is this confirmed as working OK with dynamically loaded scripts as well as unmanaged images/resources?  I'm concerned that AIUI `<base href>` only works for DOM elements and not for Javascript, and that it creates an inconsistent view of the URL space.

It's very possible that I'm thinking of an old IE bug because a long time ago I remember using `<base href>` for a while and having to get rid of it because it was causing difficulties.

EG the difference between this setup and the 5.x setup is that the `resourceUri` will reference a relative path which is not valid unless `<base href>` is taken into account.

I think that this PR at the moment will break `qx serve` because that expects to always provide it's own `index.html` so it needs to be adjusted to detect & use this index.html.

Finally, I think that the `index.html` should be deleted if `writeIndexHtmlToRoot` is turned off so that there is no possibility of it getting out of date and being accidentally used

hkollmann:
Do you have any better idea as using base tag to get an root based Installation?

johnspackman:
the `index.html` would need to load the `boot.js` from the target output directory and the `boot.js` would have to have to be written so that the `Libinfo` variable ([loader-browser.tmpl.js](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/lib/qx/tool/compiler/app/loader-browser.tmpl.js#L16)) uses paths which are relative to the top level directory, instead of relative to the output.

If `<base href>` works then that would be great because it's a very elegant and simple solution, I'm just saying that my experience is that it needs some caution and testing

hkollmann:
Ok, i will do some tests with qx serve and an new created app.
My own application works with base tag

cboulanger:
Hi, might it be useful to think about unifying/consolidating the ‘qx serve’ behavior with the writeIndexHtmlToRoot option? I.e. wouldn’t it make sense to always boot into the application if there is only one OR show the boot page if there are more as the default behavior? Any other behavior could be toggled by options but then the question wouldn’t be about “write index.html to root” because that would always happen, but about wha kind of index.html that would be - I.e. non-default boot page for a single application...

cboulanger:
I have no problem approving this PR,  but I think that eventually, `qx compile` and `qx serve` shouldn’t behave differently. That is,

A) one application-> index.html with <base> tag
B) more than one app -> boot page with logo and choice of apps to start

cboulanger:
@hkollmann wouldn’t that make the option `writeIndex...` unnecessary?

johnspackman:
2 approvals, merging thank you @hkollmann

PR #133, opened 2018-02-09
implements `qx compile --verbose`

johnspackman:
implements `qx compile --verbose` so that classes are output as they are compiled, with timings


hkollmann:
2 appovals, merged. Thanks @johnspackman

PR #134, opened 2018-02-09
fixes the scope for Rhino's script load method

johnspackman:


hkollmann:
2 approvals, merged. Thanks @johnspackman

PR #136, opened 2018-02-12
fixes a regression where compilation is broken if application names have paths

johnspackman:
also removes some old debugger statements and fixes a bug in PromisifyAll


johnspackman:
Please can someone merge this for me?

PR #137, opened 2018-02-12
fixes regression where compile can be very slow with large files

johnspackman:
fixes a regression where newer versions of "babel-plugin-transform-es2015-block-scoping" are very slow, the fix is to lock the version number at 6.15;

this also increments the version number ready for the next npm release


level420:
so as both code owners approved the PR and the PR was created to solve one of my issues with compiler: merge

PR #139, opened 2018-02-13
fixes a bug where meta data was incorrectly detecting a class as abst…

johnspackman:
…ract; this broke typescript output


hkollmann:
Thanks @johnspackman

johnspackman:
further fixes in https://github.com/qooxdoo/qooxdoo-compiler/pull/141

PR #140, opened 2018-02-14
Fixes (again) bug with slow compile

johnspackman:
Fixes permanently the issue with slow compiles for very large files, the previous solution only worked because of a bug in npm (and not consistently); the `@qooxdoo/preset-env` npm module we now rely on has been transferred to qooxdoo (see https://github.com/qooxdoo/qooxdoo-preset-env)


johnspackman:
The problem is that the `babel-preset-env` uses a later version of `babel-plugin-transform-es2015-block-scoping` which has a bug which causes a massive slow down of parsing because it uses an array scan instead of a lookup.  The `@qooxdoo/preset-env` locks `babel-plugin-transform-es2015-block-scoping` at 6.15.0 which does not have the problem (this a the solution found by someone else).

It is necessary to maintain a separate package because it is not possible to get npm to use a specific version in a nested package; yarn does support such a thing, but only if the nested package has a dependency which allows it, and in this case yarn insists on using the newer, broken module.

I did find a solution with npm, but it does not work consistently, meaning that it will revert to having multiple versions of the same plug in.

Another alternative is to collect the plugins together and ignore the preset - but that's not as easy as I first thought; I probably made a simple mistake in getting that to work, but in the end I decided that at least this way there is a very clear and simple connection to the original `babel-preset-env`, albeit with a minor tweak.

hkollmann:
So the only difference to the original preset is the version of the block-scoping plugin? Then this should be mentioned in the readme of qooxdoo/preset-env with an hint that we can use the original preset in case the babel team fixes this bug.

hkollmann:
Btw vould you add the link to to the metioned solution? So the reason why we have our own preset get‘s clearer.

johnspackman:
good point, done.  I need to file a bug report about this also, just need to get as small a test case as possible

PR #141, opened 2018-02-14
fixes bug in typescript where spaces in function names (eg unit tests…

johnspackman:
…) not copied through and constructor parameters missing

@FriedrichF


hkollmann:
Thanks @johnspackman !

Issue #138, opened 2018-02-13
Environment setting "io.ssl" not properly calculated in dependencies


johnspackman:
The tests used by qx.core.Environment have to be included in dependency calculations but this is not working properly for `qx.util.ResourceManager` and the `io.ssl` test.  

`io.ssl` is provided by `qx.bom.client.Transport` and this should be loaded before `qx.util.ResourceManager` so that it's `.defer()` method is able to use it - but it's not.  This causes SSL URLs to not work correctly on IE.

See also https://github.com/qooxdoo/qooxdoo-compiler/issues/29

PR #142, opened 2018-02-14
Boosts the dependencies of environment tests in defer

johnspackman:
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/138


hkollmann:
Thanks @kohnspackman

PR #149, opened 2018-02-19
add icontheme to templates

hkollmann:


johnspackman:
2 approvals, merging thank you @hkollmann !

PR #150, opened 2018-02-19
Fix for base

hkollmann:
this pr fixes some minor Problems with writing index.html into root:

- found a little bug in sync with filter
- only index.html should live in base dir, all other stuff from boot dir must be in application dir because it's searched relative to application dir


johnspackman:
2 approvals, merging thank you @hkollmann !

PR #152, opened 2018-02-22
move database files into target output

johnspackman:
the `db.json` and `resource-db.json` are specific to the target because of code elimination


cboulanger:
Merged, thank you @johnspackman

PR #173, opened 2018-05-03
Icon Font updates

oetiker:
* replace `testCharacters` with the qooxdoo standard `comparisonString`
* make font-map path work relative to the resource directory same as the fonts
* properly request glyphes with decimal codepoints
* use bbox glyph sizing
* add documentation

since `testCharacters` has not been documented at all, I don't see a problem with the missing backward compatibility.


oetiker:
when merging, PLEASE squash!

oetiker:
aargh ... not good yeat ... closing this one

PR #174, opened 2018-05-05
webfont fixes and better documentation

oetiker:
my second attempt

* the font 'height' is classicly associated with the size, so when giving with and height of the font, associate the size with the height
* use a resource relative path for font maps
* complain if non-existing codepoints are referenced in the fontmap
* codepoints in maps are hex, but fontkit expects them in decimal ... adapt accordingly
* updated documentation
* `testCharacters` should be called `comparisonString`


johnspackman:
full quorum, merging, thank you @oetiker !

PR #178, opened 2018-05-14
make fonts from libraries work

oetiker:
* include root dir when loading fonts, so that fonts from libraries work too

* add missing async-each dependency to package.json


johnspackman:
merging, thank you @oetiker !

PR #180, opened 2018-05-22
make --set more liberal

oetiker:
* be more liberal in accepting names for qx environment variables via the --set option. Now everything but space and = are allowed in the names.
* parse the value as js expression.


johnspackman:
3 approvals, merging, thank you @oetiker !

PR #183, opened 2018-05-28
Always use the latest npm qooxdoo version

hkollmann:
change qooxdoo-sdk version so that the newest alpha version will be used.


johnspackman:
merging thank you @oetiker !

oetiker:
it wasn't me :) ... go @hkollmann

johnspackman:
oops, sorry @hkollmann :)

PR #185, opened 2018-06-01
add a link to compile-json.doc

hkollmann:


hkollmann:
@derrell did the same direct in the repo!

PR #192, opened 2018-06-23
Result of `npm audit fix`: Reduce number of vulnerabilities from 15 to 7

cboulanger:


hkollmann:
I think you have to change Package.json too
Package-lock.json is partly generated

derrell:
Yes. Generally, you never touch Package-lock.json manually; you only manipulate Package.json, and npm manipulates the lock file.

derrell:
Yes you're right, if this is a result of nothing but `npm audio fix` then it should be good.

PR #193, opened 2018-06-23
WIP: Error/warning alert sound [NOT READY TO MERGE AS IS]

derrell:
This is the beginnings of work to notify the user if an error or warning occurs. My hope is that this will allow `--watch` to run in the background, doing its thing without the developer having a care in the world as long as it is silent, but knowing to look at the output if an alert sound is heard.

At present, when using `qx compile` (without `--watch`), this alerts for JavaScript compile errors but not for warnings such as a class not found. That is a problem, as we'd want alerts for those as well. The bigger problem, though, is that there is not currently a difference between a message displayed for a compile error, and those displayed for `--watch`'s console output, so using `--watch` generates an alert sound on each compile. Not useful.

I could use some suggestions on how to better integrate this.


johnspackman:
As messages are referenced by ID, it would be possible to associate extra information with each `msgId`, eg whether compiler warning, error etc.

One idea I had for this involves `qx serve`, in that if the web server detects a compiler error it could show a one-page website which lists the errors and warnings instead of the app.  Also, warnings could be made available via some mechanism (eg the boot loader) so the app has the opportunity to output compile warnings in the console (or a widget could be used to show a discrete notification with EG "16 warning(s)" in the top corner)

I'd like to be able to turn the sound off though

derrell:
The message ID association is good. And the distinction of whether warning or error is also good in this context, as a different sound could potentially be used in each case.

Not only should one be able to turn off the sound, but one should be able to configure which events generate a sound, e.g., I might want to have sounds for errors and warnings, where you might want them only for errors. That should probably be configurable in compile.json and on the command line. Speaking of the former... Is there an easy way for the Console module to access the parsed configuration?

johnspackman:
`qx.compiler.Console` can't see the parsed config, this is by design - the `qx.compiler.*` namespace is an API for compiling, and should not have any dependencies on `qx.cli.*` because that is only one of the way(s) in which the API could be used.

You can set the `writer` property of `qx.compiler.Console` and completely customise the output in an application specific way - `writer` is a function like this `function(stringToOutput, msgId, ...args) { console.log(stringToOutput); }`

derrell:
How about the other direction? Would it make sense for qx.compiler.Console to have an API method that allows qx.cli.* to provide configuration information to it? That API method could also be provided data (or not) by some other caller, when qx.cli.* is not used, and Console could have a set of defaults if that method is never called. Thoughts?

johnspackman:
Yes, that's what happens already for the `machineReadable` property, the CLI configures `Console` to adjust it's output.  

But as the Console is the generic, go-to class for communicating output to the app that's operating the API, it might make more sense to keep the presentation of those messages in the application (ie CLI in this case).

cboulanger:
@derrell Are you still working on this or can this be closed?

derrell:
I haven't worked on this recently. It will likely need to be done differently since the compiler is so much farther along now. I'll close this and redo some time.

PR #199, opened 2018-07-24
Catch problems in Manifest.json

cboulanger:
This will create a more meaningful error message if a `Manifest.json`'s require section has problems.


cboulanger:
Maybe it would make sense to put even more code into the `try` block, since this would catch problems with the optional settings, too...

johnspackman:
@cboulanger following on from your comment, are you going to modify this PR or is it ready for review & merge?

cboulanger:
It can be merged. We can move the other stuff into the try block later. The current change is important  because it will spot faulty manifests in contribs that have several libraries in one repo (such as ergobyte/qookery).

johnspackman:
merged, thank you @cboulanger !

Issue #156, opened 2018-03-06
qx contrib publish stopped working


cboulanger:
```
$ qx contrib publish --token=XXX
? This will increment the version to 1.0.1 and create a release of the current master on GitHub. Do you want to proceed? Yes
? Please enter a commit message: XXX
Creating tag and release 'v1.0.1' of git@github.com:cboulanger/qx-contrib-TokenField...
404 error making request POST https://api.github.com/repos/git@github.com:cboulanger/qx-contrib-TokenField/releases: "Not Found"
```

Need to investigate

hkollmann:
token is set with  qx config now

PR #201, opened 2018-07-25
Fixes #198 #156

### LINKED ISSUE #198: qx contrib publish broken
Getting error `404 error making request POST https://api.github.com/repos/git@github.com:cboulanger/qx-contrib-test/releases: "Not Found"`

cboulanger:
`qx contrib publish` uses the better maintained [OctoKit](https://www.npmjs.com/package/@octokit/rest) library instead of `github-api` now, I will switch the other commands to the newer library as time permits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@octokit/rest": "^15.9.4",
     "@qooxdoo/preset-env": "^1.0.0",
     "app-module-path": "^2.2.0",
     "async": "^2.4.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@octokit/rest": "^15.9.4",
```
johnspackman: can we also remove the dependency on `github-api` [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/ab7866fc5060f5062407975f7ef4613c13213a73/package.json#L57) also?

cboulanger: Not yet - `qx contrib update` still relies on the old library, and that still works. It's on my list but as long as there are no problems, it's low priority...

cboulanger: `qx contrib publish` was broken, that's why I had to switch, but it makes total sense to get rid of github-api completely. I might have a look at the GitHub's new, GraphQl-based API (v4), which might speed up `qx contrib update` considerably, as you can make compound queries instead of having to make several queries sequentially...

PR #203, opened 2018-07-25
DO NOT MERGE YET: Updating gitignores of skeletons

cboulanger:
Any other files that should be put into the skeletons' `.gitignore` files?


hkollmann:
db.json  and resource-db.json are changed during build - imho they should stay in the .gitignore

cboulanger:
But they are contained in the `*-output` - folders anyways, aren't they? So stating them seems redundant. Or did I overlook something?

hkollmann:
You are right..

cboulanger:
Closing because issue is already taken care of.

PR #211, opened 2018-10-03
bugfix for comment in .po Translation file

### LINKED ISSUE #210: Translation.js:190 TypeError: Cannot read property 'forEach' of null
The bugfix is
(comment.match(/[\w/\.]+:\d+/g) || []).forEach(entry => {

adrelino:
This fixes #210 .

### DIFF
--- a/lib/qx/tool/compiler/app/Translation.js
+++ b/lib/qx/tool/compiler/app/Translation.js
@@ -187,7 +187,7 @@ module.exports = qx.Class.define("qx.tool.compiler.app.Translation", {
                         entry.comments.reference = {};
                       }
                       const ref = entry.comments.reference;
-                      comment.match(/[\w/\.]+:\d+/g).forEach(entry => {
+                      (comment.match(/[\w/\.]+:\d+/g) || []).forEach(entry => {
                         const split = entry.split(":");
                         const classname = split[0];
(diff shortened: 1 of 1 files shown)


lib/qx/tool/compiler/app/Translation.js
```diff
@@ -187,7 +187,7 @@ module.exports = qx.Class.define("qx.tool.compiler.app.Translation", {
                         entry.comments.reference = {};
                       }
                       const ref = entry.comments.reference;
-                      comment.match(/[\w/\.]+:\d+/g).forEach(entry => {
+                      (comment.match(/[\w/\.]+:\d+/g) || []).forEach(entry => {
```
derrell: Looking at the code starting with line 169, it seems that `comment`, too, can be null if the line being matched contains only `#`. Maybe this should, in fact, check for comment being non-null?
```
(comment && comment.match(/[\w/\.]+:\d+/g) || []).forEach(entry => {
```

derrell: @johnspackman Can you confirm?

johnspackman: that's a good point; as they're comments, there should be no implicit assumption about formatting (quite apart from the fact that an exception causes an inexplicable failure)

PR #213, opened 2018-11-02
WIP - append qooxdoo search path

oetiker:
does  not quite work as the code is unhappy now if qx has already been included ... so we would have to check for that

### DIFF
--- a/lib/qx/tool/cli/commands/MConfig.js
+++ b/lib/qx/tool/cli/commands/MConfig.js
@@ -134,6 +134,8 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
         if (!config.libraries) {
           config.libraries = [];
         }
+        await this.__addQxPath(config.libraries,"../../../../../..");
+        await this.__addQxPath(config.libraries,"../../../../../../node_modules");
         parsedArgs.libraries.forEach(function(aPath) {
           config.libraries.push(aPath);
         });
@@ -144,7 +146,12 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
         });
       }
     },
-
+    __addQxPath: async function(arr,dir){
+        let qxDir = path.join(__dirname + dir + '/qooxdoo-sdk/framework');
+        if (await fs.existsAsync(qxDir + '/Manifest.json')){
+            arr.push(qxDir);
+        }
+    },
     /**
      * Parses the command line, and produces a normalised configuration; 
      *


lib/qx/tool/cli/commands/MConfig.js
```diff
@@ -134,6 +134,8 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
         if (!config.libraries) {
           config.libraries = [];
         }
+        await this.__addQxPath(config.libraries,"../../../../../..");
+        await this.__addQxPath(config.libraries,"../../../../../../node_modules");
```
derrell: I'm not sure this will work in all cases. If qooxdoo-sdk were installed at the time that qooxdoo-compiler is installed, I think that qooxdoo-compiler will not have in node_module the qooxdoo-sdk module because it already exists one level higher...

hkollmann: My idea is to add something like ${node_modules} into to the qooxdoo path and replace this with a searched path from node.

johnspackman: @oetiker won't this break in situations where the compile.json already refers to a global qooxdoo, eg the checked out repo?  It would mean that two different versions of Qooxdoo would be on the list of paths

@hkollmann So far I've always stayed away from any kind of variables or other processing of the `compile.json` file because of the danger of feature creep - a big issue (IMHO) with the generator's config.json was how it grew into a language with includes, namespaces, inheritance, and macros that became very hard to understand even for the simplest "hello world" project.  The mechanism for adding processing to `compile.json` is to rename it to `compile.js` and then you have code that can do anything you like when generating the config.

That's a deliberately hard line of two extremes and quite possibly not the best solution in the long run - but the advantages are that (**if** we can meet all the needs without adding processing to `compile.json`) the code base is simpler, compile.json is easier to read, and there is less temptation to add extra features (like includes).

IMHO a good long term solution would be to have some kind of global search path or default set of libraries that is set outside of the `compile.json` (ie it would be specific to each developer's machine, and by default would include the `node_modules/qooxdoo-sdk` from the qxcompiler installation directory).  Qxcompiler will only use/transpile the classes that it needs so there is very little cost in adding lots of libraries.

cboulanger: Wouldn't the best solution, then, be to ship a preconfigured `compile.js` which returns the default compile.json content, with an API call to a method that calculates the framework path?

johnspackman: @hkollmann i havn't started it, we've simply talked around it a little because it would be useful for the `qx contrib` command to be able to store things like tokens centrally also.  

@cboulanger that's possible except that it would have to work the other way around, i.e. the compile.json is read and then passed to the global `compile.js` to be edited (otherwise we would have to merge the default config with the project's compile.json, which would be complex).  I think there's room for a middle ground, where the concept of default libraries is just a simple array in a global .json file somewhere.  

In any case, there needs to be detection of duplicate libraries (based on namespace), so only one specific/predictable version is selected

oetiker: @johnspackman yes detecting duplicate mentions of libraries is the clincher ... and especially if there are different versions of the libraries ... 

I wonder if it makes sense to 're develop' this or if we could steal something from the way it is done in npm ...

cboulanger: I would suggest that any major redesign should happen in a new release. Let's fix the current system in a pragmatic way and then see how it is being used and what people would like to see changed. "Design by committee" is always a problem :-)

hkollmann:
Obsolete with this 
https://github.com/qooxdoo/qooxdoo-compiler/pull/215

PR #214, opened 2018-11-03
Source Maps: set sourceFileName explicitly

adrelino:
Set sourceFileName in babel's transform options explicitly when generating source maps. The full file path allows to distinguish between transpiled and source file even if they have the same name.

This allows chrome dev tools workspaces to save changes back to disk.
https://babeljs.io/docs/en/6.26.3/babel-core#options


adrelino:
Then there is no way to save back changes anyways and it will just use the embedded sourceContent e.g. for debugging as before.

johnspackman:
3 approvals, merging.  Thank you @adrelino !

PR #215, opened 2018-11-22
API viewer (and more)

johnspackman:
Apologies for the size of this PR, it’s grown to include more than just the headline of “API Viewer”.  This PR:

**API Viewer Meta Data** - The meta data output needed to be extended/fixed to provide the data required by the API viewer application

**Contribs-As-Applications** - Adds support for Contribs to provide applications which can be installed via the `qx contrib install <contribrepo>` mechanism

**Fixes for `qx serve`** - the web server was broken

**Fixes for the “default” application** (ie having `index.html` in the root of the output directory was broken, and `<base>` didn’t fix it)

**Persistent configuration** - it’s now possible to provide persistent configuration via the command line (see `qx config`); the most obvious uses for this are a global location of the Qooxdoo framework to use (see `qx contrib set qx.libraryPath`) and the GitHub token required by `qx contrib` (which now prompts for a token if none has been provided; see `qx config set github.token`)

**Qooxdoo library auto detection and persistent override** - if `compile.json` does not explicitly provide Qx library, then the config `qx.libraryPath` is used if set; otherwise it is detected based on `node_modules` and failing that the QxCompiler’s own Qooxdoo is used.

**Simplify Libraries** - if no `libraries` key in `compile.json`, then make a best guess including the Qooxdoo library

**JSON support for comments** - some of the configuration files were not always loaded using the JsonToAst module, which meant that comments were not supported

**CLI refactoring** - to remove dependence on a set of javascript files in `tool/cli/commands/` that existed purely to bootstrap Yargs and make the bootstrap code a class (i.e. making `index.js` very trivial)


hkollmann:
Great!!! 
Creating Server app simce to be broken. Server app works without resource dir.

johnspackman:
I've found one thing with the server and just pushed it just in case that fixes your issue - my own server apps aren't working but there's something else to do there

hkollmann:
I am just Looking at the tests. Server app is Building now.
Next problem is 
'
Run 'qx contrib list' in the root dir of your project to see which libraries are compatible.

Error: ENOENT: no such file or directory, open '/home/travis/.qooxdoo/contrib-cache.json'

    at Object.fs.openSync (fs.js:646:18)

    at Object.fs.writeFileSync (fs.js:1299:33)

    at defaultConstructor.saveCache (/home/travis/build/qooxdoo/qooxdoo-compiler/lib/qx/tool/cli/commands/Contrib.js:146:10)

    at defaultConstructor.process (/home/travis/build/qooxdoo/qooxdoo-compiler/lib/qx/tool/cli/commands/contrib/Update.js:130:14)

    at <anonymous>

    at process._tickCallback (internal/process/next_tick.js:189:7)'

lib/qx/tool/cli/Cli.js
```diff
@@ -0,0 +1,110 @@
...
+  extend: qx.core.Object,
+  
+  members: {
+    run: function() {
+      // sanitise process list parameter
+      let args = [path.relative(process.cwd(), process.argv[1])]
+
+      for (let i=2; i<process.argv.length; i++) {
+        let arg = process.argv[i]
+        if (arg === '--token' || arg === '-t') {
```
cboulanger: The token parameter has been removed, so this is no longer necessary, is it?

johnspackman: no it's not :)  I'll remove it

johnspackman:
@hkollmann I can't reproduce that, I did this:

```
Johns-MacBook-Pro:demoapp john$ rm -rf ~/.qooxdoo/
Johns-MacBook-Pro:demoapp john$ qx contrib list
You need to execute 'qx contrib update' first.
Johns-MacBook-Pro:demoapp john$ qx contrib update
Downloading cache from GitHub ...
Found 78 libraries.
Run 'qx contrib list' in the root dir of your project to see which libraries are compatible.
Johns-MacBook-Pro:demoapp john$ qx contrib list
NAME                                      DESCRIPTION                                                    INSTALLED       LATEST    COMPATIBLE
  # ... snip ...
```

@cboulanger glad you like it :)  i've quite enjoyed getting back into it

oetiker:
wow

hkollmann:
last check in breaks tests again?

johnspackman:
is this the joy of travis again?  `npm test` works fine on my machine and it's just a timeout on travis - any ideas?

hkollmann:
I ca'nt believe that this is an timeout Problem. 10 min timeout should be enough. Could it be a problem of the memory increase parameter? Could you try to remove it again?

johnspackman:
@/all now that this is passing tests I don't plan on making any further changes, and FWIW prior to this morning's push I've also been using this in production (although admittedly only for a few days)

hkollmann:
Could you publish a new npm version? Thanks!

johnspackman:
done 👍

PR #216, opened 2018-11-27
fix for qx.$$appRoot

hkollmann:
index.html is generated with qx.$appRoot instead of qx.$$appRoot.
This is a fix.


cboulanger:
Merged, thank you @hkollmann

PR #217, opened 2018-11-28
Some minor changes

hkollmann:
- remove writeIndexHtmlToRoot from template
   - fix writeIndexHtmlToRoot if only one application is configured
   - DRY: move search of qooxdoo to getGlobalQxPath in Command.js
   - if --verbose log qooxdoo path

### DIFF
--- a/lib/qx/tool/cli/commands/Compile.js
+++ b/lib/qx/tool/cli/commands/Compile.js
@@ -359,13 +359,6 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
       if (t.argv["app-name"])
         appNames = t.argv["app-name"].split(',');
       
-      // if only one app set writeIndexHtmlToRoot = true
-      if (data.applications.length === 1) {
-        let appData = data.applications[0];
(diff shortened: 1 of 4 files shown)


lib/qx/tool/cli/commands/Compile.js
```diff
@@ -362,8 +362,8 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
       // if only one app set writeIndexHtmlToRoot = true
       if (data.applications.length === 1) {
         let appData = data.applications[0];
-        if (appData.writeIndexHtmlToRoot === undefined) {
-          appData.writeIndexHtmlToRoot = true;
+        if (appData["default"] === undefined) {
+          appData["default"] = true;
         }
       }
```
johnspackman: this whole block should be deleted - the code below detects whether default has not been explicitly set (and default app only applies to browser based apps)

hkollmann: You are right

johnspackman:
thank you @hkollmann

PR #218, opened 2018-11-28
Fix app root 2: work for blank projects too.

hkollmann:
The first fix only works for projects with included boot/index.html not for blank projects.
This patch also fix the output path for serve.


johnspackman:
Merging,  thank you @hkollmann!

PR #219, opened 2018-11-30
fix relative paths in serve

hkollmann:
if you have relative paths in target.getOutputDir() app.use will not work.
Therefore use fix path to service apps

### DIFF
--- a/lib/qx/tool/cli/commands/Serve.js
+++ b/lib/qx/tool/cli/commands/Serve.js
@@ -20,7 +20,6 @@ require("qooxdoo");
 const fs = qx.tool.compiler.utils.Promisify.fs;
 const path = require('upath');
 const process = require('process');
-const http = require('http');
 const express = require('express');
 
 require('app-module-path').addPath(process.cwd() + '/node_modules');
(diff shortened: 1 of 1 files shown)


lib/qx/tool/cli/commands/Serve.js
```diff
@@ -151,11 +150,11 @@ qx.Class.define("qx.tool.cli.commands.Serve", {
         app.use('/', express.static(target.getOutputDir()));
       } else {
         app.use('/', express.static(path.join(__dirname, "../../../../../tool/cli/serve/build")))
-        app.use('/' + target.getOutputDir(), express.static(target.getOutputDir()));
+        app.use('/build/', express.static(target.getOutputDir()));
```
johnspackman: isn't this hard coding the target directory?

hkollmann: No only the url path to the server, All apps then are reached by http://localhost:8080/build/app.
If you use  target-getOutputDir() here with a traget path like '../../build' the serve will not build.
I think a hard path on the webserver is no problem.

cboulanger:
Merged. Thank you @hkollmann

PR #221, opened 2018-12-01
fixes for contib install and remove

hkollmann:
- update inquirer to version 6.2. The older one did not work in windows 
- add a npm job to create the website
- fix remove command so that the application of the contrib will be removed from compile.json. Opposite of install, there the appliccation of the contrib is added to compile.json
- fix some typos


johnspackman:

PR #222, opened 2018-12-01
write buildinfo

hkollmann:
write a buildinfo file to output directory with version numbers of compiler and all Libraries


cboulanger:
The building is currently just a value that can be processed by external scripts as-is. The information is certainly useful, but this PR would turn the file into a text file for human consumption only. I am not at all against having this information, but wouldn't it then be better to output the info as JSON so that the file stays machine-readable?

hkollmann:
Yes it  can. But with this file you have the Information in the distribution. That's the idea. Json is great, i will change to it.

johnspackman:

PR #223, opened 2018-12-01
add libinfo to db.json

hkollmann:
add lib infos to db.json. This info can be used in apiviewer to show only some trees.

Changes of apiviewer are following


hkollmann:
Needs more work

PR #231, opened 2018-12-04
Fixes #228

### LINKED ISSUE #228: qx contrib: qx contrib install should call qx contrib list implicitly
if `qx contrib list` hasn't been called previously, `qx contrib install` could do it implicitly, saving the user one step.

cboulanger:
Implicit `qx contrib update` for convenience


johnspackman:

PR #232, opened 2018-12-05
With this patch the order of libs is like the order in compile.json.

hkollmann:
This solves dependencies of external libs


hkollmann:
It's not working as expected.

PR #237, opened 2018-12-07
Update documentation on contrib dependency management

cboulanger:
I wrote down what we have discussed on Gitter and in the issues to clear up some confusion. 

OK like this?


cboulanger:
The first commit is just cosmetic, here is the actual text change: https://github.com/qooxdoo/qooxdoo-compiler/pull/237/commits/47bb0644982986feb511db5e64a6ce7618716daa

docs/cli/contrib.md
```diff
@@ -258,6 +258,31 @@ change (which is guaranteed by the semver specs), using `5.0.2 - 6.x` as the
...
+contribs (TBD). Currently, all necessary contribs will have to be declared in the
+application's `contrib.json` by first installing them with `qx contrib install
+<repo/name>`and then committing `contrib.json` to the project's repository. When
+a project is first compiled, you need to run `qx contrib install` (with no
+further arguments) to download the contribs from github.
+
+A note on NPM: Under normal circumstances, a contrib does not need to use NPM or
+maintain a `package.json` file. In particular, neither the `qxcompiler` nor
+the `qooxdoo-sdk` npm packages should be NPM dependencies of the project, since
+you should always install them globally (qooxdoo-ssk is automatically installed
```
oetiker: s/ssk/sdk/

oetiker: I would consider Installing qx compiler and qooxdoo sdk globally as gross neglichance. Just imagine you use qooxdoo on many customer projects ...  half a year or two years after shipping project-2334, the customer wants you to make a small alteration. In the meantime, both qooxdoo and the qooxdoo compiler have evolved and you have created many other projects with later versions of qooxdoo and the compiler

If you are using nvm and a local copy of the qooxdoo compiler and SDK, you can now easily modify and rebuild the project, whereas with a globally installed compiler and SDK you may have to do major surgery on the project just to get it to build again, not to speak of all the potential regressions and required testing ... 

So please, do NOT suggest to install stuff globally ... this will only lead to major headaches down the line.

hkollmann: We should'nt suggest anything for this special point - let it be part of user decision. We can describe both.
BTW a little typo: qooxdoo-ssk should be qooxdoo-sdk

cboulanger: Fair points. I have updated the documentation accordingly. However, this is not the place to discuss the pros and cons of installing the compiler globally, this should be discussed in the main section on installing - which I have linked to. Maybe we should update that section?

PR #239, opened 2018-12-09
Some fixes to qx create

cboulanger:
This fixes a number of lint warnings in `qx create`, refactors out certain file paths as static constants, and adds the automatic  listing of skeleton types to the help text/documentation. Also, previously the application namespace wasn't requested interactively.


johnspackman:

PR #242, opened 2018-12-12
Adds support for nested namespaces in qx create

### LINKED ISSUE #238: qx create does not handle non-global namespaces
ATM, `qx create` can only deal with global namespaces. Needs to be rewritten to translate dotted namespaces into the appropriate structure of the `source` directory.

cboulanger:
I.e. a contrib named "foo.bar.baz" is now allowed and will put the source files into source/class/foo/bar/baz/. Resources will be in source/resource/foo.bar.baz/
Fixes #238


oetiker:
what is the idea behind NOT putting resources in resources/foo/bar/baz ?

cboulanger:
#243 is relevant here. Waiting until it is resolved.

lib/qx/tool/cli/commands/Create.js
```diff
@@ -270,14 +277,19 @@ qx.Class.define("qx.tool.cli.commands.Create", {
...
             }
           } else if (stats.isDirectory()) {
-            if( part === "custom" ) part = values.namespace;
-            let newTargetDir = path.join(targetDir,part);
-            fs.mkdirSync(newTargetDir);
+            let newTargetDir = targetDir;
+            // replace "custon" with namespace, creating namespaced folders in the "class" dir, but not anywhere else
+            let parts = (part === "custom")
+              ? (targetDir.endsWith("class") ? values.namespace.split(/\./) : [values.namespace])
+              : [part];
```
hkollmann: if you change this to
let parts = (part === "custom") ? values.namespace.split(/\./) : [values.namespace])  : [part]; 

it will work for resources too

cboulanger: Thank you. Can you confirm that the current compiler expects resources in the nested directory structure (see #243)? If yes, I could implement your suggestion and #243 could be closed.

hkollmann: all qxl contribs uses nested directory structure for resources.

cboulanger: If I do `qx create foo.bar.baz -I && cd foo.bar.baz && qx serve` with the proposed change and the latest qxcompiler, the icon of the default demo application is not found.

cboulanger: bildschirmfoto 2019-01-02 um 13 54 50

cboulanger: So there is still some issue there. Sorry that I cannot look at it, I have to work on something entirely unrelated currently...

hkollmann: That's a Problem of serve!

cboulanger:
Ups, sorry for this commit mess. Anyways, closing in favor of #264

PR #248, opened 2018-12-16
Here are some fixes to get demobrowser to work:

hkollmann:
1. Extend Compile.js to through some of the maker events so that you can listen to them in compile.js.
2. set qx.$$appRoot in template if not set
3. Targets:  Some fixes for getScriptPrefix. This was not handled correct everywhere
4. BuildTarget.js: Enable path mapping for resources.


cboulanger:
Two approvals, merging. Thank you @hkollmann

PR #249, opened 2018-12-16
some fixes for the templates

hkollmann:
add external css resource to mobile/Manifest.tmpl.json
remove hybrid target from compile.tmpl.json


johnspackman:
thank you @hkollmann !

PR #251, opened 2018-12-19
add auto install in case of missing contrib

hkollmann:
I case of missing contrib call qx contrib install to install the contrib

### DIFF
--- a/lib/qx/tool/cli/commands/Compile.js
+++ b/lib/qx/tool/cli/commands/Compile.js
@@ -472,6 +472,10 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
           });
         }
       });
+
+      if (!data.libraries.every(libData => fs.existsSync(libData + "/Manifest.json"))) {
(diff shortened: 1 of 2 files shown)


lib/qx/tool/cli/commands/Compile.js
```diff
@@ -472,6 +472,10 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
           });
         }
       });
+
+      if (!data.libraries.every(libData => fs.existsSync(libData + "/Manifest.json"))) {
+        await (new qx.tool.cli.commands.contrib.Install({quiet:true})).process();
```
cboulanger: Maybe inform the user about implicit installation unless a "quiet" flag has been set?

hkollmann: we have no message in auto update and list either. Should'nt we inform the user in this case too?

hkollmann: Done

cboulanger:
2 approvals (fast-track procedure), merging. Thank you @hkollmann !!

Issue #247, opened 2018-12-13
Missing css file in mobile app


eda2z:
Executing the following commands:

qx create mtest --type mobile -I
cd mtest
qx compile

results in the following error:

2018-12-13 09:25:21.098 Error: ENOENT: no such file or directory, open '/srv/http/qx6/mtest/source/resource/mtest/css/custom.css'

The resulting app runs but it looks "ugly".  Am I missing something?  I am using the latest qx compiler on arch linux.

cboulanger:
I can confirm this. It's odd. I first thought it was related to #243, but the resource folder is completely missing for some reason. I am looking into it.

cboulanger:
I am not getting the error if I use the current master. Which version is displayed when you do `qx` without any arguments?

eda2z:
qooxdoo-compiler v0.2.25.  I just did an 'npm -g update' this morning.  If I do a 2nd 'qx compile' without an intervening 'qx clean' then I don't get the error but the displayed result still looks 'strange'

hkollmann:
2 runs of compile are necessary for the first time - the generated css files are not reckognized in the first run.
What's ugly? Do you have a picture?

eda2z:
I was probably being harsh using the word ugly.  I guess I was expecting a "smoother" default theme.
![mtest](https://user-images.githubusercontent.com/6683633/49962939-4f360280-fedc-11e8-8a98-efcb1277f8f4.png)

hkollmann:
Your are right. The custom.css file is not loaded.
Please try to insert an externalResource in your manifest.json 

`  "externalResources": {
        "script": [
         ],
        "css": [
          "your_app/css/custom.css"
        ]
   }
`
We should do this during create for mobile app

cboulanger:
@hkollmann can you fix the [mobile skeleton](https://github.com/qooxdoo/qooxdoo-compiler/tree/master/tool/cli/templates/skeleton/mobile)?

hkollmann:
It‘s on the list

Von meinem iPad gesendet

eda2z:
Since I am actually trying to convert several mobile apps from 5.0.2 to the compiler, I copied over the files from old_app/source/resource/old_app/css to new_app/css and updated the Manifest.json.  But nothing happened so I'm guessing I placed the css directory in the wrong location.

hkollmann:
css or scss files? I mobile app the css files are autogenerated into css dir. This is not changed.
[...]

PR #253, opened 2018-12-19
.gitignore files are ignored by npm pulish. So they do not exist in n…

hkollmann:
…pm install package


hkollmann:
This will fix https://github.com/qooxdoo/qooxdoo-compiler/issues/247

PR #255, opened 2018-12-20
With this fix it is allowed to have scss files in resource path

hkollmann:


cboulanger:
Two approvals (fast-track) - merging. Thank you, @hkollmann 🎄

PR #256, opened 2018-12-28
Added more events to Compile.js.

hkollmann:
catch "made" event in Serve.js to log hint when all making is done and the apps are avaiable
update to qooxdoo-server 6.0.0-alpha to get promise support


lib/qx/tool/compiler/targets/Target.js
```diff
@@ -133,6 +133,16 @@ module.exports = qx.Class.define("qx.tool.compiler.targets.Target", {
 
   },
 
+  events: {
+    /**
+     * Fired after all enviroment data is collected
+     *  application {qx.tool.compiler.app.Application} the app 
+     *  enviroment: {Object} enviroment data
+     */
+    "checkEnviroment": "qx.event.type.Data"
```
cboulanger: Typo: should be „Environment“

lib/qx/tool/compiler/targets/Target.js
```diff
@@ -413,6 +423,8 @@ module.exports = qx.Class.define("qx.tool.compiler.targets.Target", {
 
         for (var name in environment)
           configdata.environment[name] = environment[name];
+          
+        t.fireDataEvent("checkEnviroment", { application: application, enviroment: configdata.environment});
```
cboulanger: Typo: should be "checkEnvironment"

johnspackman:

PR #257, opened 2018-12-28
cache must be reread from disk after call to list to get the changes

hkollmann:


johnspackman:

Issue #236, opened 2018-12-07
Compiler should output state of compilation into json file


cboulanger:
(from Gitter discussion)
The Compile class should routinely output state into an object, and when it changes, that state is optionally written out to a .json file. That would allow external code to watch state easily, and would provide the basis for some kind of visual feedback later (eg serve a “please wait, compiling..” page or “you have errors in your code” page).

hkollmann:
With https://github.com/qooxdoo/qooxdoo-compiler/pull/256 compiler fires an event after all is done.
This event is caputured in serve and print's a log to show that app is avaiable

hkollmann:
Or we could start serve after the whole work is done...

cboulanger:
@hkollmann Thanks, this is very useful. (I would, however, capitalize "Application" in the "application ready" message or maybe replace it by "Compilation done"). 

I am in favor of logging "Web server started, please browse to http://localhost:8080" to the console AFTER this message.

cboulanger:
@hkollmann Very good, thank you, just what we need!

hkollmann:
done with https://github.com/qooxdoo/qooxdoo-compiler/pull/263

PR #263, opened 2019-01-01
Let's the server start after all building is done

hkollmann:
see https://github.com/qooxdoo/qooxdoo-compiler/issues/236


johnspackman:

Issue #19, opened 2017-06-02
Constant compile / watch mode


johnspackman:

johnspackman:
done, several versions ago now :)

Issue #259, opened 2018-12-31
Add json schema for compile.json & Manifest.json


cboulanger:
For validation and convenient autocomplete in IDEs, it would be good to have a JSON schema for `compile.json` and `Manifest.json`. This would also induce us to keep documention of these files up-to-date.

cboulanger:
@johnspackman

cboulanger:
https://json-schema.org/specification.html

thron7:
There is also an old schema spec for the old `config.json` in the repo:

https://github.com/qooxdoo/qooxdoo/blob/master/tool/data/config/config_schema.json

Might help as another resource.

cboulanger:
To be implemented by #365

Issue #235, opened 2018-12-07
Compiler should check qooxdoo-range and warn/abort if not matched


cboulanger:
The compiler should check the `qooxdoo-range` setting in `Manifest.json` (with a fallback to `qooxdoo-versions` for v6) and either
1) output a warning if the qooxdoo version used does not match this range or
2) abort compilation unless forced with `-f/--force`

I vote for option 2)

hkollmann:
From gitter discussion:
What's about adding a about a block 
`neededVersion:` { 
"compiler": "0.2.16", 
"libraries": 
{ 
"apiviewer": "0.1.7", 
"qx": "6.0.0-alpha-20181116-263f2cc9" `}` 
to compile.json? 
Each line should be semver compared to lib and compiler and raise an error if version did not match?

cboulanger:
Shouldn't it be in `Manifest.json`? It's a description of actual dependencies which should be enforced, not a compile-time setting.

hkollmann:
But your description describes it as compile-time setting?

cboulanger:
Yes, but a structural one that the user cannot/should not change, unlike compile.json which allows user modification. What do you think @johnspackman?

hkollmann:
ok, i think this are 2 problems:
1. Compiler needs a checked qooxdoo-sdk version
2. Application needs checked versions of qooxdo-sdk and contribs.

johnspackman:
Everything is a library so IMHO the minimum version of dependencies should be in the libraries `Manifest.json` ... the catch is that `Manifest.json` does not specify dependencies currently, because these are specified in `compile.json` and `contrib.json`.

@cboulanger IMHO you're right that this would normally go into `Manifest.json` *if* we were doing automatic dependency management, the problem is that we only do this in `compile.json` at the moment.

I'm not against automatic dependency management, but I am a little cautious to avoid the "dependency hell" and the huge amount of code and effort that went into solving it for npm.  Perhaps if we declared that the list of dependent libraries/contribs is flattened into a single list before compilation, and that any conflicting versions must be resolved manually that would be ok?  But maybe that's how `npm` started out...  

An elegant solution could be to have `Manifest.json` declare what contribs and versions are required by the contrib (this can optionally include the `qooxdoo-sdk` with a version, otherwise it would be assumed by the compiler).

The contrib/library's `compile.json` and `contrib.json` would be completely ignored when that contrib is being added to some other application/compilation.  

[...]

PR #268, opened 2019-01-06
issue 235: Compiler should check qooxdoo-range and warn/abort if not matched

hkollmann:
This PR adds following features:
- new section in Manifest,json:
```
 "requires": {
      "qooxdoo-sdk": "^6.0.0-alpha",
      "qooxdoo-compiler": "^0.2.40",
       "qooxdoo/qxl.versionlabel": "^10.1.1",
       "qooxdoo/qxl.formdemo": "^0.1.1"
   }
}
```

- Compiler do a semver check of required contribs. This check is be done for all installed contribs.
 Error is thrown if version do not match. Should this be overrun by parameter?
- Compiler do a semver check for sdk and compiler version.  Error is thrown if version do not match,  Should this be overrun by parameter?
- install.js adds all required contribs to contrib.json. This is be done recursive.
- install.js adds intalled contrib to requires section of the current project. Should this be disabled by parameter?
- "qooxdoo-range" is be marked as deprecated. Instead of info.qooxdoo-range requires.qooxdoo-sdk should be used.
- Changed documentation of Manifest,json to reflect this.
- implemented issue 244: allow qx install library@version
- implemented issue 196: qx contrib install: more helpful message if no compatible release #19

### DIFF
--- a/lib/qx/tool/cli/commands/contrib/Remove.js
+++ b/lib/qx/tool/cli/commands/contrib/Remove.js
@@ -74,7 +74,7 @@ qx.Class.define("qx.tool.cli.commands.contrib.Remove", {
       }
       
       await this.removeApplication(repo_name);
-      await this.__deleteRequieredFromManifest(repo_name);
+      await this.__deleteRequiredFromManifest(repo_name);
 
       // read libraries array from contrib.json
       let data = await this.getContribData();
@@ -102,7 +102,7 @@ qx.Class.define("qx.tool.cli.commands.contrib.Remove", {
 
     },
 
-    __deleteRequieredFromManifest: async function(repoName) {
+    __deleteRequiredFromManifest: async function(repoName) {
       let contribJson = await this.getContribData();
       let libraryJson = contribJson.libraries.find(data => data.repo_name == repoName);
       if (!libraryJson)
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,7 @@ node_js:
 - "stable"
 # - lts/* # once we support on-the-fly transpilation
 
-before_script:
-  - echo ${GH_TOKEN}
-  
+
 stages:
   - name: test
   - name: deploy
--- a/test/test.linux.sh
+++ b/test/test.linux.sh
@@ -4,8 +4,11 @@ qx create myapp -I --type server -v || exit $?
 cd myapp
 qx compile -v --clean || exit $?
 node source-output/myapp/myapp.js || exit $?
-qx config set github.token ${NPM_TOKEN} || exit $?
-qx contrib update --search -v|| exit $?
+
+#bootstrap: new format is not avaible and env vars are not working in pr
+exit 0
+
+qx contrib update -v|| exit $?
 qx contrib list    -v|| exit $?
 qx contrib install oetiker/UploadWidget -v || exit $?
 qx contrib install cboulanger/qx-contrib-Dialog -v || exit $?


cboulanger:
I am in favor of deprecating `qooxdoo-range` which we just introduced in favour of this new manifest section. But the question in my [comment](https://github.com/qooxdoo/qooxdoo-compiler/issues/235#issuecomment-451647802) from #235 is stil relevant, I think: Are we checking *library* versions, *npm package* versions, or *contrib* versions?  The `Manifest.json` defines a *library*, not a contrib (which can contain several libraries), so one would expect it to require libraries in a specific version. Which *contrib* these libraries are shipped with seems to be a secondary issue. Contrib versions are defined by the tags of GitHub releases. This is my own design, maybe not the best, but a pragmatic choice. I just want to make sure we are not making decisions which will create hairy problems later.

cboulanger:
Thanks for all the work, by the way!

hkollmann:
I implemeted the check against the version in manifest.json.

Von meinem iPad gesendet

cboulanger:
Wouldn't it make sense to keep the name of the contribs out of the `Manifest.json` and only refer to library namespaces? I.e. `"qx"` instead of  `"qooxdoo-sdk"`, `"qx.tool.compiler"` instead of `"qooxdoo-sdk"`, `"qxl.versionlabel"` instead of `"qooxdoo/qxl.versionlabel"`? contrib names and versions  would only be stated in `contrib.json`. This would seem to be a much clearer separation of concerns. It would also allow to swap contribs while keeping libraries. Although it would introduce ambiguity about *which* library is meant by the info in the `"requires"` section - but that would be solved by `contrib.json`, the equivalent of `package.json` in npm land. On the other hand, one could consider `Manifest.json` as `package.json` and `contrib.json` as `package-lock.json`. Thoughts?

hkollmann:
1. You can only install contribs, not libraries. requires is a list of all install contribs. So if we list namespaces in requires we can not auto install all contribs.
2. Manifest.json desribes only one library. A contrib with more then one library has not Manifest.json in the root. Instead there should be a qooxdoo.json. I did not found any contrib which uses this structure - so i did not test this feature.

IMHO Manifest.json is equivalent to package.json in npm land and conrtrib.json as package-lock.json.

The general idea is to 

- create a new app
- add a single contrib. Have all depended contribs auto installed.

commit only Manifes.json. After checkout a qx compile runs qx contrib install implicit and this installs all needed stuff.




After

cboulanger:
I agree with all of your intentions. I just think that `Manifest.json` and `contrib.json` should be kept separate, and that `requires` should mirror `provides`, inasmuch as it should only refer to libraries. Otherwise, it seems to me, we are mixing separate conceptual and architectural levels. 

I think this becomes visible because you have to make  special ad-hoc rules for "qooxdoo-sdk" and "qooxdoo-compiler" in your syntax. Namespaces feel much more natural (as they stay within the conceptual frame of `provides`) and they don't require any knowledge about the particular way we think about contribs. 

The contrib system was meant to be built *on top* of the core, we should be able to exchange it with something completely different. That's why I would like to keep contrib-related stuff in `contrib.json` event though then we don't have a separation between version and lock file (yet) that achieves the separation between `package.json` and `package-lock.json`.

Does that make sense to you?

johnspackman:
:+1: for the repo names instead of library namespaces - the other problem with relying on namespaces only is that we would make it impossible for anyone to fork a library repo (or have conflicting namespaces by chance).  

Although it would be nice to have the simplicity of referring to namespace, I don't think it's possible to make it work in reality and IMHO it's much simpler and user friendly than imposing rules for namespaces (eg "gh." prefix).

While repos are currently assumed to be github, we could in the future support some other syntax eg "gitlab:user/repo" or even plain old "http://mydevserver:8080/repo".

The only hitch then is that the compiler should detect situations where multiple libraries are included with conflicting namespaces.

johnspackman:
thanks for all the work that's gone into this BTW, this is another of many great new additions!

cboulanger:
Hm, I wonder why we need to touch `Manifest.json` at all, which means integrating an still experimental tool chain into the core library.  As I have said to Henner privately on Gitter, as much as I like our current contrib system, I don't know how well it will perform in practice when used  by many people - as opposed to us early adopters. Maybe we will have to introduce massive changes later or someone will write a better contrib system. At the moment, we can still exchange the contrib system and nothing needs to be changed in core. That's why I would prefer to use `contrib.json` for all of the dependency management, and rather introduce a new file such as `contrib-lock.json` if we want to separate *version range* and *actual* version.

I am not fundamentally opposed to this PR, I am just worried that we hardwire too many ad-hoc solutions and microprotocols that will be hard to change later when we know how contribs are actually used on a bigger scale... That's why I didn't make an attempt to solve this quite complicated problem. But I am certainly thankful for Henner's work!

hkollmann:
>           Will it accept "Console" (with capital "C")?

Sorry, typo.

hkollmann:
BTW if we add the namespace info to contrib-cache.josn it should be possible to use Namespace in requires. Compiler has to check and warn about double namespaces. 

requires should be 

`"requires": {
      "qooxdoo-compiler": "^0.2.40",
      libraries: {
          "qx": "^6.0.0",
          "qxl.versionlabel": "^10.1.1",
         "qxl.formdemo": "^0.1.1"
     }
 }
  
`
then.

lib/qx/tool/cli/commands/Compile.js
```diff
@@ -522,19 +521,19 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
         if (markers) {
           markers.forEach(function(marker) {
             var str = qx.tool.compiler.Console.decodeMarker(marker);
-            t.warn(data.classFile.getClassName() + ": " + str);
+            Console.warn(data.classFile.getClassName() + ": " + str);
```
johnspackman: should be lower case `c`

lib/qx/tool/cli/commands/Compile.js
```diff
@@ -567,7 +619,7 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
       if (type == "hybrid")
         return qx.tool.compiler.targets.HybridTarget;
       if (type == "typescript")
-        throw new qx.tool.cli.Utils.UserError("Typescript targets are no longer supported - please use `typescript: true` in source target instead");
+         qx.tool.cli.Utils.UserError("Typescript targets are no longer supported - please use `typescript: true` in source target instead");
```
johnspackman: missing `throw` and `new`

lib/qx/tool/cli/commands/contrib/Publish.js
```diff
@@ -159,7 +155,7 @@ qx.Class.define("qx.tool.cli.commands.contrib.Publish", {
       if( ! fs.existsSync(manifest_path) ) {
         throw new qx.tool.cli.Utils.UserError("Cannot find Manifest.json - are you in the project directory?");
       }
-      let manifest = jsonlint.parse( fs.readFileSync(manifest_path,"utf-8") );
+      let manifest = JSON.parse( fs.readFileSync(manifest_path,"utf-8") );
```
johnspackman: This should be `qx.tool.compiler.utils.Json.parseJson` because `JSON.parse` does not support comments

lib/qx/tool/cli/commands/contrib/Update.js
```diff
@@ -203,7 +202,7 @@ qx.Class.define("qx.tool.cli.commands.contrib.Update", {
               let data = qooxdoo_data.data;
               if (typeof data == "string") {
                 try {
-                  data = jsonlint.parse(data);
+                  data = JSON.parse(data);
```
johnspackman: also requires `qx.tool.compiler.utils.Json.parseJson` because `qooxdoo.json` is hand edited

lib/qx/tool/cli/commands/contrib/Update.js
```diff
@@ -247,10 +246,10 @@ qx.Class.define("qx.tool.cli.commands.contrib.Update", {
               }
               // retrieve compatible qooxdoo versions
               let data = manifest_data.data;
-              // todo see above
+              // @todo check if the method can return JSON to save parsing
               if (typeof data == "string") {
                 try {
-                  data = jsonlint.parse(data);
+                  data = JSON.parse(data);
```
johnspackman: also `qx.tool.compiler.utils.Json.parseJson`

lib/qx/tool/cli/commands/contrib/Install.js
```diff
@@ -68,164 +72,236 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
...
+
       return await qx.tool.compiler.utils.Json.saveJsonAsync("compile.json", compileJson);
     },
 
     __loadFromContribJson: async function() {
-      // read libraries array from contrib.json 
+      // read libraries array from contrib.json
       let contrib_json_path = process.cwd() + "/contrib.json";
-      let data = fs.existsSync(contrib_json_path) ? jsonlint.parse( fs.readFileSync(contrib_json_path,"utf-8") ) : { libraries : [ ] };
+      let data = fs.existsSync(contrib_json_path) ? JSON.parse(fs.readFileSync(contrib_json_path, "utf-8")) : {libraries : []};
```
johnspackman: This could be `qx.tool.compiler.utils.Json.parseJson` because `JSON.parse` does not support comments - although this is debatable given than `contrib.json` is typically edited by `qx` only, and when written out any comments would be lost.  

It *is* important for other `.json` files though (see review below)

hkollmann:
Thanks @johnspackman! I just want to eliminate jsonlint because we should have only one Json parse lib.

cboulanger:
@derrell Your assumptions are correct, and that is why I am hesitant to put the contribs into `Manifest.json`. I know this makes everything much more complicated - Henner's current design is much more straightforward. 

On Gitter, I have discussed with Henner whether a cached contrib->lib mapping would solve our problems:

1. `qx contrib install x` downloads contrib x. 
2. It analyzes its `Manifest.json` and `contrib.json` and assembles a list of dependencies  (namespaces and their versions given in `requires`), and of imports (namespaces from `provides` in the version given by `info.version`).
3. `qx` and `qx.tool.compiler` (which I would prefer over `qooxdoo-sdk` and `qooxdoo-compiler` would be treated separately).
4. It downloads each contrib and loads each individual `Manifest.json` and `contrib.json` recursively, extending the list of imports and dependencies.
5. When all contribs have been loaded, the list of dependencies is matched with the list of imports. If namespaces that are listed in the dependencies are missing in the list of imports, or the versions do not match, compilation stops and the user is informed.

I know this would make everything much more complicated, but I think it's a more future-proof way.

hkollmann:
2. needs help of qooxdoo-cache.json. qx update must generate a mapping of namespace -> contribs.
3. qx.tool.compiler is not a needed namespace. Therefore i would seperate it. 
5. is done in start Phase of qx compile. No other possibility - the compiler loads all libraries on startup. After this the checks can be done.


With the changes to qooxdoo-cache.json the implentation is simple. 
My current design can not handle the case where more than one Namespace is in the contrib. 

The only unsolved problem is if 2 contribs publish the same namespace. This is something qx install must deal with. IMHO ist should throw an error if the same namespace is imported twice and give a hint to remove the first

derrell:
What does this actually mean?

> My current design can not handle the case where more than one Namespace is in the contrib.

Each library by definition has a different namespace, so if multiple libraries are in a contrib, they will all have different namespaces. Does this then mean that the namespace prefix must be the same for all libraries? If I have library `derrell.mystuff.api` and library `derrell.mystuff.app` it would work, but if I had `derrell.api` and `example.app` in the same contrib, it would not?

hkollmann:
the current implementation uses the contrib name in requires - this is not working for the ergobyte/qookery with more then one library in the contrib.
So must switch to namespace in the requires section.

cboulanger:
Ok, seems like we are converging on a good solution. 

1. `qx contrib update` collects additional information that maps contrib versions to the namespaces that are imported by all the libraries (===`Manifest.json` files) contained in the individual contrib version (===release). 
2. `qx contrib install` modifies `contrib.json` with the current behavior, plus modifies `Manifest.json` adding the versioned namespaces imported by the installed contribs. The dependencies of these contribs are contained in their respective `Manifest.json` (libraries) and `contrib.json` (contribs).
3. `qx compile` checks whether the compiler version matches, plus if all of the library version match. 

Did I summarize this correctly? 

It is fun to think about something as complex and challenging, and we should get it right.

cboulanger:
What makes somewhat difficult is that the concept of "library" isn't fully consistent. A library is basically a `Manifest.json` file and can provide several namespaces. It is not uniquely identified by one namespace or even a name, only by its location in the filesystem, which is arbitrary.

hkollmann:
The manifest.json can only provide one namspace!
` "provides": {
    "namespace": "test",
    "encoding": "utf-8",
    "class": "source/class",
    "resource": "source/resource",
    "translation": "source/translation",
    "type": "application"
  },`

cboulanger:
@hkollmann You’re right!

hkollmann:
To 1.: During implementation i found that a mapping is not necessary. It's enough to add the namespace to manifest section in qooxdoo-cache.json and deliver this to contrib.json.

docs/compiler/Manifest.md
```diff
@@ -95,9 +96,8 @@ This is a sample file:
...
-     `contrib`: `needed_version` where `needed_version` is a semver compatible version description.
-     
-     Special handling for:
-     * qooxdoo-sdk: This checks the version of the used qooxdoo sdk.
-     * qooxdoo-compiler: This checks the version of the used qooxdoo compiler.
+*   **requires**: all required dependencies.
+  * qooxdoo-sdk: Version of compiler app can be compiled with
+  * libraries: List of needed namespaces. Special handling for `qx`, used as version of qooxdoo sdk
+    * `qx`: `needed_version` version of qooxdoo sdk, mandatory for contribs. Otherwise it will be not listed in the catalog.
+    * `contrib`: `needed_version` where `needed_version` is a semver compatible version description.
```
cboulanger: shouldn't this be `namespace`: `needed_version`?

lib/qx/tool/cli/commands/contrib/Update.js
```diff
@@ -263,25 +264,25 @@ qx.Class.define("qx.tool.cli.commands.contrib.Update", {
               // qooxdoo version @deprecated
               var qx_versions = data.info["qooxdoo-versions"];
               // @deprecated
-              var qx_version_range = data.requires["qooxdoo-sdk"] || data.info["qooxdoo-range"];
+              var qx_version_range = (data.requires && data.requires.libraries && data.requires.libraries.qx) || data.info["qooxdoo-range"];
```
cboulanger: Interesting! I did not know that in JavaScript `("blue" && "black" && "green")` will return `"green"` and not boolean  `true`....

lib/qx/tool/cli/commands/contrib/Install.js
```diff
@@ -171,8 +169,14 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
         } else {
           data.libraries.push(library_elem);
         }
+        if (writeToManifest && !manifest.requires.libraries[namespace]) {
+          manifest.requires.libraries[namespace] = info.version;
```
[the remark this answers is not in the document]
cboulanger: (that's the real beauty of semver ranges)

lib/qx/tool/cli/commands/contrib/Update.js
hkollmann: I learned this yesterday: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators

lib/qx/tool/cli/commands/contrib/Remove.js
```diff
@@ -103,9 +103,14 @@ qx.Class.define("qx.tool.cli.commands.contrib.Remove", {
     },
 
     __deleteRequieredFromManifest: async function(repoName) {
```
cboulanger: typo: `s/Requiered/Required/g`

lib/qx/tool/cli/commands/contrib/Install.js
```diff
@@ -151,6 +151,9 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
       let library_elem = null;
       for (let {namespace, info, path:filename} of release_data.manifests) {
         // does the repository name already exist?
+        if (!namespace) {
+          throw new qx.tool.cli.Utils.UserError("Tag 'namespace' not found. Please run qx contrib update!");
```
cboulanger: maybe implicit update instead? although this will lead to infinite recursion until the cache is updated for the first time, so maybe this should be done in a later PR.

hkollmann: Implicit update will not work - place to Deep in the program. And after Manual update this is fixed for ever.

johnspackman:
Just as a follow up on some of the comments above, I think that it's important to be clear that namespaces are largely meaningless now - they **only** exist to be able to identify a library *internally during compilation* by some kind of unique, human readable identifier, and they have no bearing on any of the class names provided by that library.

That uniqueness could easily be exchanged for some kind of system-generated unique ID, one which is only used internally during the compilation process ... in other words, the user's concept of namespace could (and perhaps should) disappear entirely.

This works because the compiler discovers classes wherever they are, and so "my.package.MyClassOne" and "my.package.MyClassTwo" could be in two different libraries.

(There is one exception to this - the "qx" library is still special, but there is only one place this is needed in the compiler and this could easily be abstracted out)

Making dependencies between libraries (aka contribs) is easy provided the dependencies refer to the repo, and not some arbitrary "namespace" - (a) this allows people to fork repos and not change the namespace or package names, and (b) namespaces are arbitrary and prone to conflict because of the historical practice of one-word namespaces.

Sorry if this is now stating the obvious - I'm reading the conversation from yesterday and catching up.  What I've written here conflicts with what @derrell and @cboulanger were discussing yesterday, but AIUI @hkollmann 's implementation is based on repos (in which case correct IMHO)?

cboulanger:
@johnspackman 

> Making dependencies between libraries (aka contribs) is easy provided the dependencies refer to the repo, and not some arbitrary "namespace" - (a) this allows people to fork repos and not change the namespace or package names, and (b) namespaces are arbitrary and prone to conflict because of the historical practice of one-word namespaces.

It is important to see that, at least in the current implementation,  "libraries" !== "contribs". A library is only uniquely identified by its `Manifest.json`, which `provides` a namespace, which - your are right, is largely arbitrary. Which might or might not be a design mistake of the current implementation. A "contrib" can contain any number of "libraries", which makes it so confusing.

That is why I am so fiercely insisting on NOT hardcoding the contrib names in the core, because some still fundamentally unsolved design questions. We need a firm, and shared, understanding of the architecture: what "library" and "contrib" actually mean. I don't think we're there yet.

oetiker:
just an additional point ... libraries/namespaces have no versions ... only contribs have versions ... so, for now, I would stick with contribs and maybe add the possibility for contribs to tell (or is this already done?) which namespaces they provide ... I don't think the @hkollmann s point about 'maybe someone writes a much cooler contrib system in the future' is valid, since even if this should happen, then datastrcutures can be converted easily ... as long as we do not allow for programming instructions at this point in the system, then all reamains totally open to be supplanted by something cooler ... if that should ever happen ... but lets cross that bridget when we reach it ...

hkollmann:
The new version can tell which namespaces a contrib provides

hkollmann:
Let's try a definition:
contrib: a repository in Github which contains one or more libraries.
library: A piece of code to extent qooxdoo. Description in Manifest.json. A library can require other libraries.
namespace: The name of the library to reference it.

hkollmann:
example of contrib with more than one library is found here https://github.com/ergobyte/qookery

cboulanger:
@oetiker 
> just an additional point ... libraries/namespaces have no versions ... only contribs have versions

There are several misunderstandings here that show that we need to work this out in more detail! I am at fault here, too, because it seems I never really documented the contrib system enough. 

- Libraries DO have versions, because a library is tied to a `Manifest.json`, and the manifest has a `info.version` property. 
- Contribs have a version, but that is only a *convention*, because the version is the name of the *release tag*, which could be "blabbla" or "qooxdoo_rulez" or what have you.  Of course, that wouldn't make sense. The real *versions* are those of the *libraries* contained in that contrib, which could be completely different from the version tag of the contrib. This might be a stupid design, but that is how it is currently.

oetiker:
on the server side I work with perl a lot, and perl has a very active namespace system ... often a single package contains code which lives in different namespaces ... but it all belongs to the same package and has the same version ...instead of re-inventing things it might be worth looking at how this problem is solved in perl ...

in any event, also in perl,  dependencies are defined to packages and not namespaces or libraries ...

cboulanger:
@oetiker And how are the packages "namespaced"? like npm, with a single global namespace?

cboulanger:
And re: package, not library dependencies - this was the way it was before this PR (without recursive auto-install). `contrib.json` specified the contrib dependencies, not `Manifest.json`(which is tied to the *library*). There are good arguments for putting dependencies into `Manifest.json`, but they should not be at the *contrib* level, because that is one level higher (remember, contribs contain libraries === `Manifest.json` files)

hkollmann:
We could easily remove the usage of the Namespace and use the name tag instead. Namespace was used in the old python build.

cboulanger:
Maybe the solution could be to require all contribs to have a `qooxdoo.json` file (currently, only qookery has one), which contains the contrib dependencies. That would be the right layer since it would be placed at the root level of a repo. This way, we leave `Manifest.json` alone and put the information where it belongs...

hkollmann:
But then we have another file to maintain… And with the current implentation you can add only one library from a contrib which contains more than one.

cboulanger:
I know it is an additional file (but one that we already maintain and document). But it would clear up all the confusion with namespaces and libraries and be a simple "contrib requires contrib" type of relationship. It would save us a lot of conceptual headaches and remain firmly within the current implementation without side effects on the core.

cboulanger:
I think that the following config files would make sense and would provide the right separation of concerns:

1. library -> `Manifest.json`
2. compilable application -> `compile.json`
3. contrib which can contain several applications and libraries and can have dependencies -> `qooxdoo.json`

oetiker:
> @oetiker And how are the packages "namespaced"? like npm, with a single global namespace?

yes there is a global namespace and often packages contain files that integrate into different namespaces ... for example if a package provides plugins for a bunch of other packages then each plugin will be in the plugin namespace of the respective package

oetiker:
hmmm how about we talk features first and then implementation? it seems to me that we are flying a tad to low at the moment ... because when we agree on the feature set we need, then maybe the implementation becomes clear.

cboulanger:
@oetiker I fully agree - we should identify the problem we need to solve and only implement the features that are absolutely necessary to solve the problem. Otherwise we risk to over engineer and to hardwire stuff that doesn't solve an urgent problem.

cboulanger:
Ok. Restart brain. New try.

Let's look into a `Manifest.json` of a contrib:
```
{
  "info": {
    "name": "My very special contrib,
    "summary": "",
    "description": "",
    "homepage": "",
    "license": "MIT license",
    "authors": [ ],
    "version": "0.2.0",
    "qooxdoo-range": "^6.0.0-alpha"
  },
  "provides": {
    "namespace": "contrib.special.my",
    "encoding": "utf-8",
    "class": "source/class",
    "resource": "source/resource",
    "translation": "source/translation",
    "type": "contribution"
  }
}
```

There is a `provides.type` property with the value "contribution". In theory, this should give us the possibility to make a rule that there can be *only one* of this type of Manifest in the repository, that it must be in the top directory of the repo and that it contains all information (including dependencies) on the contribution. This would make `qooxdoo.json` redundant, and we could remove it.

cboulanger:
More specifically:
- a `Manifest.json` of type "contribution" would allow a "requires" section that refers to contributions
- we would integrate the information contained in a `qooxdoo.json` file. Here is [an example](https://github.com/ergobyte/qookery/blob/master/qooxdoo.json). Instead of specifying `"contribs"`, it would specify `"libraries"`, i.e. `Manifest.json` files which must be of `provides.type` type "library" and NOT "contrib"!!!

cboulanger:
a qookery-type contrib would then have a Manifest that looked like this:
```
...
  "provides": {
    "type": "contribution",
    "libraries": [
      { "path": "./qookery" },
      { "path": "./qookeryace" },
      { "path": "./qookerycalendar" },
      { "path": "./qookerymaps" },
      { "path": "./qookerymobile" },
      { "path": "./qookeryrichtext" }
    ]
  }
```
@johnspackman does Manifest.json also contain information on Applications or is that in `compile.json` only?

derrell:
Can we back up further?

Assume no implementation. There's no such thing as Manifest.json,
qooxdoo.json, or anything else beyond building one's own app or library
locally. What are the *requirements* regarding what we're trying to do? The
goals? What are we trying to accomplish?

I'll start:

   - From a user's perspective, be able to install a needed library easily,
   and associate it with an app in development
   - From a developer's perspective, package multiple related libraries
   together into a wrapper that is installed as a whole, and then provides all
   of its libraries as options for the user to make use of.
   - ???

Let's fill this out with the high-level requirements. We can then look at
possible implementations (starting with what we have implemented,
presumably) to see how they can accomplish those requirements.

Derrell

hkollmann:
User perspective:
qx create test
cd test
qx contrib install qooxdoo/qxl.widgetbrowser
qx serve

If qooxdoo/qxl.widgetbrowser needs other contribs - what it actually do - should they be loaded automatic.

hkollmann:
Developer perspective:

qx create contrib
cd contrib
qx contrib install qooxdoo/qxl.widgetbrowser
qx contrib publish

User can use this contrib without further work. The new generated contrib must store the dependency of qooxdoo/qxl.widgetbrowser automatically so that another user can use the contrib. See above.

hkollmann:
@derrell: Is that what you mean?

derrell:
Yes, that's a beginning. Good. And from that you also presumably want to do `qx contrib update` to retrieve ones that have changed, right? So is it libraries that get updated, or wrappers (contribs) that get updated? Or is it the libraries that get updated, but that requires also updating the wrapper, so that whenever a library version number changes, there is a corresponding change to the wrapper version number?

Or would we like to just have version numbers on libraries, and somehow the wrapper knows what library version numbers it contains so that `qx contrib update` can work?

johnspackman:
I'm utterly confused as to why this is an issue - i must be missing something here!

@cboulanger said:

> It is important to see that, at least in the current implementation, "libraries" !== "contribs". 

Really?  Why is this?  Or why was this changed?  Because a contrib has always been just a library (not multiple libraries), they are the same thing -- the only thing we call a contrib is when that library can be detected by `qx contrib`, ie it's a repo with `qooxdoo-contrib` tag.

(BTW an application is also *just* a library)

I don't see any reason for this to change, and the only explanation I've seen offered is that a contrib needs to offer multiple namespaces -- except the "namespaces" concept is no longer used...

> A library is only uniquely identified by its `Manifest.json`, which `provides` a namespace, which - your are right, is largely arbitrary. Which might or might not be a design mistake of the current implementation. 

Do you mean the `namespace` property in `Manifest.json` (which is now irrelevant, not just arbitrary)?  Or the fact that the library/contrib package names are defined by the user and can have conflicts?

> A "contrib" is just a release of a repository on GitHub and can contain any number of "libraries", which makes it so confusing.

Well, a library/contrib can have lots of packages, but IMHO it is only confusing if you try and define the what the top level package is and rules about whether overlaps can occur or how to handle conflicts.

> That is why I am so fiercely insisting on NOT hardcoding the contrib names in the core, because some still fundamentally unsolved design questions. 

I don't understand what you mean about hardcoding contrib names - when are we doing this?

> We need a firm, and shared, understanding of the architecture: what "library", "namespace", and "contrib" actually refer to. I don't think we're there yet.

Ironically, I think we already have a working solution, and its the one we've had for ages - the only issue I can see is that attempting to add contribs via namespace is fraught with problems and will not work without a registry, and will make it hard for anyone to fork someone else's library/contrib.

> Recursive dependencies / `Manifest.json` / `compile.json` / etc

Given that library === contrib === application, and a library/contrib/application has a `Manifest.json` which lists  library/contribs it requires (in the form of a list of repos with version numbers) why can this not be resolved recursively into a flat list of repos?  And then those repos checked out, so that when the compiler runs, it scans all the checked out directories and compiles?

(Note that at no point in any of this does anything care about what the namespaces might be)

> `qooxdoo.json`

This is the exception about one repo to one library - it just acts as a further redirection, which is useful only for finding contribs for `qx contrib list`.  I can't recall whether this has namespaces in it, but if it does it can delete them because a dependency on a repo-with-multiple-librarys would be expressed as `"username/reponame/libraryOne" : "^1.0.0"`

hkollmann:
@derrel: qx contrib update is not for updating a contrib. It's purpose is generate a list of contribs on GitHub. This list contains all projects with a qooxdoo-contrib tag. 
The project must have a Manifest.json file with it's decription

cboulanger:
@johnspackman The issue is confusing, and it is hard to follow because the PR has started with one particular problem (checking dependencies) and has morphed into something completely different. Plus, this PR has started to solve complex problems that later we realized led into the wrong direction.  @hkollmann and I already figured out that the whole namespace stuff led nowhere and I think in the next iteration, Henner will throw that out. So the whole idea of matching libraries with versions has been abandoned, at least by me. We returned to the original idea of having dependencies only between contribs. 

The difference we had was about where to put dependencies in a situation where there are potentially multiple `Manifest.json` or `compile.json`. This is an edge case since in most cases, a contrib contains only one library. But by design, it CAN contain several, so we have to account for that, and then the question is, which of the `Manifest.json` files defines the dependencies. Initially, I have argued to use `qooxdoo.json` since it is already established by the docs that there will be only ONE of these in a repository. But since the Manifest has a type property, it can be marked as the "contribution" Manifest and is therefore well-suited to contain the dependency (despite my earlier resistance) and can replace `qooxdoo.json` if we want to.

cboulanger:
@johnspackman So did I understood you correctly that the `"provides.namespace"` entry in `Manifest.json` is not needed by the compiler *at all*? I.e., that it can removed after we drop support for the python toolchain, and can be removed from the `qx create` script & templates?

hkollmann:
I think the "username/reponame/libraryOne" : "^1.0.0" is the right solution for multiple libraries in one contrib. That's the needed hint, thanks @johnspackman.

So  IMHO i should return to the first attempt with a requires section in Manifest.json like this
`
 "requires": {
      "qooxdoo-sdk": "^6.0.0-alpha",
      "qooxdoo-compiler": "^0.2.40",
       "qooxdoo/qxl.versionlabel": "^10.1.1",
       "qooxdoo/qxl.formdemo": "^0.1.1",
       "ergobyte/qookery/qookeryace": "^0.5.0"
   }
`
And i will remove namespace from qx create & templates if @johnspackman agrees.

cboulanger:
Yes, this sounds good.

And should we integrate the information contained in `qooxdoo.json` into the top-level `Manifest.json` and remove `qooxdoo.json`?

hkollmann:
Idea is good - but please NOT with this PR! 
We can prepare another PR for this!

johnspackman:
> We have been talking at cross purposes the whole time. [...] Once we entangle that mess and get our terminology straight, there is probably not much left that we disagree about.

That's what I'm thinking :)

> We should very definitely clean this up. A JSON schema (#259) will be a good way of looking through all the fields and discard them if they are not needed.

👍

Issue #270, opened 2019-01-10
Allow to install a contrib library identified by its path within a contrib repository 


cboulanger:
Currently, `qx contrib install` works per-repository, i.e. all of the contrib libraries in this repository are installed into `contrib.json`. We want to support the syntax `qx contrib install owner_name/repo_name/path/to/dir-containing-manifest` to be able to specify dependencies between singe contrib *libraries*, not between contrib *repositories*.

PR #271, opened 2019-01-10
Fix for #270

cboulanger:
With this change, the following is possible:
```
qx contrib install ergobyte/qookery/qookeryace --verbose
>>> qooxdoo version:  6.0.0-alpha
>>> Repo has already been downloaded to /Users/cboulanger/Code/qooxdoo-compiler/foo/contrib/ergobyte_qookery_v0.5.0. To download again, execute 'qx clean'.
>>> Updating already existing config.json entry 'ergobyte/qookery/qookeryace'.
>>> Installed Qookery Ace (ergobyte/qookery/qookeryace, v0.5.0)
>>> No application to install.
>>> Done.
```

### DIFF
--- a/lib/qx/tool/cli/commands/contrib/Install.js
+++ b/lib/qx/tool/cli/commands/contrib/Install.js
@@ -36,7 +36,7 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
   statics: {
     getYargsCommand: function() {
       return {
-        command: 'install [repository]',
+        command: 'install [contrib_uri]',
         describe: 'installs the latest compatible release of a contrib library (as per Manifest.json). Use "-r <release tag>" to install a particular release.',
         builder: {
           "release" : {
@@ -79,7 +79,7 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
 
       if ( repos_cache.list.length === 0 ){
         if (!this.argv.quiet) {
-          console.info("Updating cache...");
+          console.info(">>> Updating cache...");
         }
         this.clearCache();
         // implicit update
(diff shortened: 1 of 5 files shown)


lib/qx/tool/cli/commands/contrib/Install.js
```diff
@@ -88,28 +88,31 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
         repos_cache = this.getCache().repos;
       }
 
-      if (!argv.repository) {
+      if (!argv.contrib_uri) {
```
johnspackman: does this break existing contrib.json files?  I'm not against the change, just wondering if there's any steps that need to be taken

cboulanger: No, I made an effort to make this backwards-compatible. The `contrib.json/libraries[]` maps get an additional `uri` property but stay the same otherwise. `qx contrib install` without arguments works just as before.

lib/qx/tool/cli/commands/contrib/Install.js
```diff
@@ -88,28 +88,31 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
...
      
-      // has a repository name been given?
-      if (!argv.repository) {
-        throw new qx.tool.cli.Utils.UserError("No repo name given");
+      // has a contrib_uri name been given?
+      if (!argv.contrib_uri) {
+        throw new qx.tool.cli.Utils.UserError("No contrib resource identifier given");
       }
-      let repo_name = argv.repository;
+      // currently, the contrib uri is github_username/repo_name[/path/to/repo].
```
cboulanger: this should be `/path/to/dir-containing-manifest` not `path/to/repo/`, but I can correct that later...

johnspackman:

PR #280, opened 2019-01-18
Update documentation

cboulanger:
This PR updates the qooxdoo-compiler documentation to the latest changes. 

I have changed one additional thing based on dissatisfaction with our recent recommendation to use the namespace `contrib.github_user.project_name`. I now think that `user_name.project_name` is really enough for the namespace (or any other two-part namespace, for that matter). The `contrib` part doesn't add any information and looks ugly and is a drag to type all the time (`contrib.myname.myproject.mysubnamespace.MyClass`!!!).


cboulanger:
@hkollmann I know but `@version` is much nicer and more intuitive, thanks for implementing that!

PR #281, opened 2019-01-19
Fix lint errors in qx contrib list

cboulanger:


cboulanger:
Closing since @hkollmann has fixed lint errors in an upcoming PR.

PR #284, opened 2019-01-22
with this PR qooxdoo-compiler lints without errors

hkollmann:


lib/index.js
```diff
@@ -17,11 +17,12 @@
...
 require("qooxdoo");
-if (qx.tool === undefined) 
-  qx.tool = {};
+if (qx.tool === undefined) {
+  qx.tool = {}; 
+}
 require("./qx/tool/cli.js");
 require("./qx/tool/compiler.js");
 
-debugger;
```
johnspackman: this is really useful (almost essential)

hkollmann: no Problem...

hkollmann:
without indenting the eslint fixing is ugly. Therefore i suggested to use the same indenting as pretty in qooxdoo. See https://github.com/qooxdoo/eslint-config-qx/pull/2

johnspackman:
If our eslint settings enforce an indenting style, then we have to 100% guarantee that all of our edits comply with that linting style - I think it's very likely that one day we will run eslint as part of travis, and that setting will mean that we will potentially forever be fighting our editor(s).  

Also, I have yet to see a beautifier that always gets the layout of object literals consistently right (by "right" I mean easily readable), especially in large object literals which often have special layout requirements.

It's arguably a bug in eslint that it breaks when reformatting without this setting, because obviously when asked to `--fix` code it's dumb to simply delete all indentation - but I do not think that is a reason to turn on strict styling rules like whitespace.

When I agreed to that PR, I did mention that I thought that it would enforce these rules but couldn't reproduce it -- I don't know why I couldn't reproduce it before, but this is exactly the reason why I removed that setting in the first place.

> Therefore i suggested to use the same indenting as pretty in qooxdoo

Does the generator's lint job enforce indentation to such a degree?  I didn't think it was that strict.

hkollmann:
Generator lint did not care about indenting. That's done by generate pretty.

IMHO we should have run indenting one time to have a nice looking. After this we could remove it in .eslintrc. 
Could you agree with this?

johnspackman:
Well, the thing is - I don't think that it really needs beautification that much!  Yes, there's some inconsistency in some places, but this is an arbitrary set of rules put together by a group of people somewhere else that we're just deciding to adopt "as is".

From my point of view, my editor (Eclipse) has lead to a lot of the indentation that's in use in `qx.tool.compiler.*` and that's nearly half the files in this PR; I havn't had as much involvement in `qx.tool.cli.*` but there's still a huge impact in code that doesn't seem to have been causing anyone any issues so far.  

So my question is what benefit will this change in indentation bring, and what cost it will be to maintain it in the future.  When the indentation level changes to something else, then that needs to be maintained - it wouldn't be ideal to have part of an object literal using one style and then to suddenly change.

My second question would be why are eslint's preferences for layout somehow better/prettier/more readable/etc?  It feels like an arbitrary decision, with no value and possible down sides.

cboulanger:
I don't fully understand what the state of the discussion is but I do think we should disable all rules that concern indenting which will cause a change when running `eslint --fix`. If we have a separate command that does what `generate pretty` did, that's fine (since it is optional), but those rules shouldn't be enforced for commits.

hkollmann:
any suggestions how to proceed? Could you live with one time run of ident to fix curly? Or should i revert the changes of Compiler part?

johnspackman:
I don't think it is valid to do this to *any* part of the code - the wholesale reformatting is simply not necessary and provides no benefit.  For the compiler parts, I ran `--fix` and then had to fix the indentation bugs caused by curly - it's a pain but quite quick to do, and curly has done the hard work of adding the `{}`.  

I realise you've already done a chunk of work on this that will be impossible to unpick, so I'm happy to do that if you'd like - I can probably make some time to do it today?

We could drop the `curly` rule too, and although I'd be in favour of that, it's an existing Qooxdoo standard and IMHO it's not appropriate to make that decision here.

johnspackman:
approved, merging thank you @hkollmann

PR #286, opened 2019-01-23
Trim library output

johnspackman:
excludes libraries which are known to the compiler but not actually used by the application from the compile process; this reduces the data written to `boot.js` and will avoid processing unneeded resources and webfonts


cboulanger:
Thanks @johnspackman!

PR #287, opened 2019-01-24
Auto detect `qx.$$appRoot` and make absolute

johnspackman:
Makes the boot loader convert `qx.$$appRoot` into an absolute path and auto-detect the `qx.$$appRoot` - this allows a web page to add just `<script src="/path/to/boot.js"></script>` and wherever that is, the app will load.

### DIFF
--- a/lib/qx/tool/cli/commands/Compile.js
+++ b/lib/qx/tool/cli/commands/Compile.js
@@ -153,8 +153,6 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
             type: "boolean",
             default: true
           }
-
-          
         },
         handler: function(argv) {
           return new qx.tool.cli.commands.Compile(argv)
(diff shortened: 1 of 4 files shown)


johnspackman:
sorry, this isn't quite ready for review yet - please hold off

hkollmann:
qxl.apiviewer and qxl.demobrowser needs to be tested after changes to qx.$$appRoot. They use this variable to access their ressources.

johnspackman:
ok; the variable still exists, it's just auto-detected and made absolute.  but i'll check them out

johnspackman:
@hkollmann the api viewer works, the demobrowser doesn't - but nothing to do with `qx.$$appRoot` it looks like the repo is out of date?  the class `qxl.demobrowser.demo.data.store.Flickr` is missing for example...

johnspackman:
this is ready for review now - i've tested it against api viewer OK, let me know if/when there is a demo browser to test against.

The potential for breaking is that `qx.$$appRoot` was an optional, relative path whereas now it's a fully calculated path eg `http://www.example.com/myapp/`; the demo browser code only seems to append to this variable so it should be fine.

PR #288, opened 2019-01-25
with this watch is working again

hkollmann:


johnspackman:
merging, thank you @hkollmann

PR #289, opened 2019-01-25
Add version check

hkollmann:
npm has this nice little feature to give a hint when there is a new version available on npm. With this have the same for qx!


hkollmann:
2019-01-25 14_01_30-c__windows_system32_cmd exe

Global install is not suggested!
And without the intrusive message it will be easily overseen. And the message comes maximum once a day.

cboulanger:
Merged, thanks @hkollmann

PR #293, opened 2019-01-28
Remove warning since the contrib consumer cannot do anything about it…

cboulanger:
…. This needs to be enforced via JSON schema.


johnspackman:

PR #299, opened 2019-02-04
Fixes a bug where the top namespace was wrongly compared 

cboulanger:
... with the given full namespace


johnspackman:

PR #303, opened 2019-02-12
Fix broken `qx compile --watch`

johnspackman:
fixes a regression introduced in https://github.com/qooxdoo/qooxdoo-compiler/pull/302


cboulanger:
Merging to avoid broken npm installation, assuming @hkollmann's consent. Thanks @johnspackman !

PR #306, opened 2019-02-12
fixes bug in merging environment variables for compile

johnspackman:


cboulanger:
Fix for blocking bug (at least for me), fast track, assuming approval of @hkollmann, thank you very much for resolving this @johnspackman !

PR #308, opened 2019-02-14
Fix env vars regression for targets

johnspackman:
fixes regression where target environment vars are ignored;
adds unit tests for regression;
fixes issue where command line `--listen-port` is ignored for `qx serve`;
better message when qx serve port is in use already


cboulanger:
Urgent bug fix, fast track, assuming @hkollmann' approval. Thank you @johnspackman !

PR #312, opened 2019-02-16
it's not allowed to watch the same directory more than once

hkollmann:


johnspackman:

PR #313, opened 2019-02-16
make the check for boot.js less strict. Here is a use case:

hkollmann:
<script>
    [
       "settings.qxrpc" + location.search
	 , "${appPath}boot.js"
    ].forEach(function(src) {
	var script = document.createElement('script');
        script.type = "text/javascript";
	script.src = src;
	script.async = false;
	document.head.appendChild(script);
    });
  </script>


johnspackman:

PR #315, opened 2019-02-19
use qooxdoo sdk as contrib

hkollmann:
To use the qooxdoo-sdk as contrib 2 problems have to be solved:

- use Manifest.json instead of version.txt to read qooxdoo version.
- remove dependency from tool/data/cldr and use an npm based cldr solution.

This PR adresses both.


johnspackman:
merging, thank you @hkollmann

PR #319, opened 2019-02-21
move version to 0.3.0-alpha as discussed on gitter.

hkollmann:
change deployment to npm to use 0.3.0-alpha.yyyymmdd.hhnn as version number


cboulanger:
Merging, thank you @hkollmann !!

PR #320, opened 2019-02-22
Don't update package.json due to new qooxdoo-sdk version

derrell:
When checking out qxcompiler from git, running `npm install` should not change any committed files. This prevents the embedded `npm update qooxdoo-sdk` from changing the `package.json` file.


johnspackman:
merging, thank you @derrell !

Issue #279, opened 2019-01-18
Drop "contrib" from "qx contrib XXX"


cboulanger:
The more we use the "qx contrib" commands, the more it seems to me that we might as well drop the `contrib` part of `qx contrib (update|list|install|publish|...)`. We use the commands to install *library dependencies*, some of which are "contributions" from the community, others are libraries developed by the core team. It's really not only about "contribs" any more. So to save some typing and also to make the commands more "general", we could simply get rid of the "contrib" part. 

What do you think?

cboulanger:
This would be my vision for the library infrastructure before we go into beta:

- we drop the "contrib" semantics in the CLI, in the code  and in the documentation completely, and speak of "libraries" only. This will simplify things quite a bit, I think.
- `qx install` is decoupled from `qx update/list`:
  - `qx install` installs code from arbitrary URIs, using adapters for different top-level domains (GitHub, GitLab, etc.) that know how to resolve that URI to one from which a ZIP containing the whole library 
can be downloaded. 
  - `qx update/list` is the *discovery service* that has a list of the libraries published on GitHub. If `qx install` can get stuff from anywhere, we can limit our discovery service to GitHub. If `qx install` encounters a shorthand URI, it resolves it using the cache built by `qx update/list`.

I think this would make qooxdoo's package/plugin mechanism more coherent. I could invest a weekend or two on this. Don't know when I get the chance, though.

derrell:
+1

cboulanger:
Superseded by #369

PR #321, opened 2019-02-24
Fix message shown when a library is not found.

cboulanger:
The current message is ambigous, might be understood to mean that the user should "try to install..."
Also removing reference to "contrib" in preparation of #279.


hkollmann:
full approval, merged. Thanks @cboulanger!

PR #325, opened 2019-03-05
boot.js should not use urls with //

oetiker:
when boot.js assembles the urls for the components/parts to load it can end up with double-slashes in the url. the built in webserver of qx does not mind, but for example the aoihttp server from python does not like such urls at all .. this patch fixes it.

As @johnspackman mentioned on gitter

> the policy I’ve tried to take is that directories always get ‘/‘ appended in advance, so that “” is valid to prepend to a path

I have not found the code responsible for building FontLoading urls which would end up with double slashes when changing resourceUri too ... maybe @johnspackman can help there.

### DIFF
--- a/lib/qx/tool/compiler/app/loader-browser.tmpl.js
+++ b/lib/qx/tool/compiler/app/loader-browser.tmpl.js
@@ -1,6 +1,6 @@
 (function(){
 
-if (!window.qx) 
+if (!window.qx)
   window.qx = {};
 
 qx.$$start = new Date();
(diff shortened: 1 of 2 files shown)


cboulanger:
Cannot merge because of an error 
```
[Error] Cannot load script https://github.githubassets.com/assets/frameworks-b740280f.js. Failed integrity metadata check.
[Error] Unhandled Promise Rejection: Error: evaluating module ./github-bootstrap-10ed91b2.js: could not resolve ./frameworks.js
	(anonyme Funktion) (github-bootstrap-10ed91b2.js:1:792)
	promiseReactionJob
```

PR #327, opened 2019-03-07
in #325 oetiker removed the trailing / in loader-browser.tmpl.js to g…

hkollmann:
…et rid of //. No it's necesarry to append / to resource.

### DIFF
--- a/lib/qx/tool/compiler/targets/BuildTarget.js
+++ b/lib/qx/tool/compiler/targets/BuildTarget.js
@@ -81,7 +81,7 @@ module.exports = qx.Class.define("qx.tool.compiler.targets.BuildTarget", {
       var targetUri = t._getOutputRootUri(application);
       var appRootDir = this.getApplicationRoot(application);
       var mapTo = this.getPathMapping(path.join(appRootDir, this.getOutputDir(), "resource"));
-      var resourceUri = mapTo ? mapTo : targetUri + "resource";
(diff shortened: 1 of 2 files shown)


cboulanger:
Maybe instead of removing the slash, it would be better to add a `.replace(/\/\//,"/")` where necessary.

johnspackman:

oetiker:
note that I did NOT mess with the resources path:

https://github.com/qooxdoo/qooxdoo-compiler/pull/325/files?utf8=%E2%9C%93&diff=unified&w=1#diff-48adb84e3faed9722c679affa66153fdL59

I think there are probably more problems ahead ... I did some experiments, but found that I ended up with double slashes in resources paths!

johnspackman:
@hkollmann do you have an example of why you needed to add the trailing '/' in?  I've just tried this and it looks that resourceUri does not need to have a trailing "/" because `qx.util.ResourceManager.toUri` always adds one

hkollmann:
Yes, i have. Just an example with 

  "externalResources": {
    "script": [
      "js/zip.js"
    ]
  }
[test.zip](https://github.com/qooxdoo/qooxdoo-compiler/files/2947450/test.zip)


Running gives:  Cannot load script test/../resourcejs/zip.js

zaucker:
See also my example in the chat.

PR #332, opened 2019-03-13
speeds up the compilation process when using `qx compile --watch`

johnspackman:
if you have a lot of apps (eg 7 or 8) this makes a huge difference (eg 4.5 secs down to 0.5)


hkollmann:
Tests are not running?

cboulanger:
Merged, thank you @johnspackman and looking forward to the speed-up!

PR #334, opened 2019-03-15
Ensure Console is defined

rad-pat:
This is giving me an error in the current compiler version, this is hopefully the fix


rad-pat:
Other fix is available

johnspackman:
@rad-pat thanks for this, I didn't realise you were on the case when I put my one in!

rad-pat:
No problem, I announced the issue and then just figured that I could probably just fix it instead of moaning about it!! :)

PR #338, opened 2019-03-20
fix for "Unable to correct case for library in" in windows systems

hkollmann:


cboulanger:

PR #339, opened 2019-03-23
add boot.js and  ${preBootJs} to template

hkollmann:


johnspackman:

PR #340, opened 2019-03-23
RootDir in lib is absolute now, so check for requires libs fails. This fix the issue

hkollmann:


johnspackman:

PR #342, opened 2019-03-26
This fixes a problem with installing sublibraries

cboulanger:


johnspackman:

PR #345, opened 2019-03-29
Improvements to qx publish support for multiple libraries in a repository

cboulanger:
- create index file with `--create-index`
- all Manifest.json files are updated with new version number
- ability to select main library which determines version (so that version numbers in the libraries do not have to be manually updated)


docs/cli/contrib.md
```diff
@@ -200,20 +204,35 @@ repository in one of the following ways:
...
+
+You can (re)generate the `qooxdoo.json` file by using 
+`qx contrib publish --create-index`. The command will automatically search for all
+`Manifest.json` files in the repository and ask you to select the 
+main library if you have more than one.
+
+Note that all libraries in the repository must have the same version number,
+because dependencies are managed and checked on the level of the repository, not
+of the library. When you `qx contrib publish` the repository, the versions of
+all libraries will be set to the one of the main library.
```
oetiker: hmmm this seems an odd restriction ... can you remind me why the contrib system must be able to handle this case. Why is it not enough to allow to have multiple libraries in a single git repo, but appart from that treate them as separate contribs ?

cboulanger: The problem is - as we discussed earlier - that it would make the contrib system much more complex if you allowed multiple libraries to have divergent version numbers. You would have to add a separate lookup to the cache to store which version of which library is in which release of which repository. The rule that all libraries need to have the same version as the release name solves a lot of problems that we would have to address otherwise. We can think about a different system later, but for the moment it seems to make sense to me to keep it simple.

johnspackman:

PR #347, opened 2019-04-03
Fix bug with qx contrib update where Manifest path in root was malformed

cboulanger:
This happened only in repos that have a qooxdoo.json and was therefore not noticed yet.


cboulanger:
Closing because of conflicts.

Issue #182, opened 2018-05-23
Add qx contrib upgrade


cboulanger:
`qx contrib upgrade` should install the latest compatible version of all installed contribs. This is basically the same `qx contrib install <packagename>` for each contrib.

hkollmann:
qx install - without package - do this for the configured version.

cboulanger:
Yes, but to force upgrade you would need to delete the 'contrib' directory, which doesn't seem very convenient. A short script command could do just that and would be a future-proof API if things get implemented differently internally.

cboulanger:
@johnspackman As `qx upgrade` doesn't work currently and can be confused with the new `qx contrib upgrade`, I think it would be better to remove `qx upgrade` and replace it later with `qx migrate` in order to avoid confusion.

cboulanger:
Implemented by #349.

PR #349, opened 2019-04-04
Implements #182

cboulanger:
adds qx contrib upgrade


johnspackman:
merging thank you @cboulanger

Issue #350, opened 2019-04-04
Allow "qx contrib install my/library@<commit  hash>"


cboulanger:
In order to make development of contribs easier, we should allow to reference a commit directly, bypassing the contrib cache altogether. This way, code trees can be referenced without having to wait for releases to be made (and registered in the cache).

oetiker:
could we  make this 'tree-ish' instead of commit hash ?

cboulanger:
True, makes more sense!

hkollmann:
Done with #351,

cboulanger:
Thanks. I am reopening because I saw this:
https://help.github.com/en/articles/getting-permanent-links-to-files
It says: "You can put any identifier that can be resolved to a commit in the URL, including branch names, specific commit SHAs, or tags!"
Maybe this can really simplify our dowload URL resolution issue.

cboulanger:
That is, one URL template `https://github.com/<user_name>/<repo_name>/blob/<id>/` would cover both release tags and commit hashes, and we can get rid of the currently hardcoded "tree/" prefix .

cboulanger:
Although `/bob/` doesn't work because we need the ZIP file.

oetiker:
https://github.com/qooxdoo/qooxdoo-compiler/archive/<tree-ish>.zip

hkollmann:
Could we close this now?

cboulanger:
Yes, it’s implemented now. Closing.

PR #351, opened 2019-04-05
Implement install from tree hash

cboulanger:
Implements #350

### DIFF
--- a/lib/qx/tool/cli/commands/contrib/Install.js
+++ b/lib/qx/tool/cli/commands/contrib/Install.js
@@ -124,10 +124,12 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
       this.__data = await this.getContribData();
 
       // install library/libraries
-      if (qx.lang.Type.isString(id) && id.startsWith("tree/")) {
-        await this.__installFromTreeHash(uri, id, true);
-      } else {
+      if (qx.lang.Type.isString(id) && id.startsWith("v")) {
+        // we have a release
         await this.__installFromRelease(uri, id, true);
+      } else {
+        // arbitrary path into the code tree
+        await this.__installFromTree(uri, id, true);
       }
 
       // save data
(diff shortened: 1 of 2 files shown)


oetiker:
does this mean I can now use any tree-ish after the @ ? tag/branch/hash

cboulanger:
so far, only "tree/" is recognized, but in principle this should can be done with path-like id if it can be differentiated from a version. maybe this is why it might make sense to require "v" as a version prefix for all releases, everything else would be "tree-ish"?

cboulanger:
This line is differentiating "releases" from "non-releases", we could change it accordingly:
https://github.com/qooxdoo/qooxdoo-compiler/blob/0b6eabda22940e8067786f99fc5c568e348efaec/lib/qx/tool/cli/commands/contrib/Install.js#L127

cboulanger:
Or we merge my code and then you change in a way that you think would be better? I have no particular preference here....

cboulanger:
We also need tests and documentation, but I don't have more time than for what I've done this morning.

oetiker:
https://stackoverflow.com/questions/4044368/what-does-tree-ish-mean-in-git

at least on the git cli client 'tree-ish' is a 'thing' so I would assume that if you put it in the right place that all would happen automatically without requir3ement on yourside to have special handling

cboulanger:
I see (but don't fully grasp the details). As the current way allows to address any arbitrary state of the code, it is enough for my purposes. I am open for anything that would make it more flexible & more "git-ish", but I am afraid I cannot put the time into implementing it myself ATM. Sorry.

cboulanger:
now it no longer depends on the "tree/" prefix, but you can use any path into the code tree. This requires naming releases with a "v" prefix. That's already a requirement but one that I thought could be relaxed.

cboulanger:
Sorry, didn't run `npm test` so there is a regression. Fixing it right away...

cboulanger:
In any case, I haven't added other "tree-ish" expressions than "tree/<hash>". Feel free to do that here: 

https://github.com/qooxdoo/qooxdoo-compiler/blob/0b39d79bf1c5efff02a0c7e9891e1c54116e730a/lib/qx/tool/cli/commands/contrib/Install.js#L402-418

PR #352, opened 2019-04-05
Renames "qx upgrade" to "qx migrate"

cboulanger:
goal is to to differentiate from "qx contrib upgrade" and to better
describe what the command is doing


johnspackman:

PR #354, opened 2019-04-05
BUGFIX: it is necessary to wait for __updateInstalledLibraryData. 

hkollmann:
Otherwise the dependend lib will not be written to contrib.json


cboulanger:
I will merge this despite the failing tests, assuming @johnspackman's approval, then fix the problem in the test and provide a new PR.

cboulanger:
Thanks @hkollmann, forgetting `await` before async functions has bitten me so often, I must be more careful.

PR #356, opened 2019-04-06
move compiler to babel 7

hkollmann:
With this we can use new language features like destructing inside the code.


hkollmann:
Will reopen this after Import works again.

lib/qx/tool/compiler/ClassFile.js
```diff
@@ -286,7 +286,7 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
               filename: t.getSourcePath(),
               sourceMaps: true,
               "presets": [
-                [ require.resolve("@qooxdoo/preset-env"), { "modules": false } ]
```
johnspackman: i realise this PR is closed, but were you able to check that qooxdoo/preset-env is no longer required?  There was a bug which had a big performance impact, details at https://github.com/qooxdoo/qooxdoo-preset-env/blob/master/README.md

hkollmann:
I will reopen it after tests running again!

lib/qx/tool/compiler/ClassFile.js
hkollmann: They have reworked this in babel 7:

 if (block.body) {
      const declarPaths = this.blockPath.get("body");
      for (let i = 0; i < block.body.length; i++) {
        addDeclarationsFromChild(declarPaths[i]);
      }
    }

hkollmann:
With babel 7 you can use destructuring assignment

      let obj = {
        t1: 1, 
        t2: 2
      };   
      const obj1 = {t1, ...obj};

PR #357, opened 2019-04-08
Additions to qx list

cboulanger:
This adds a couple of new options to `qx contrib list`:

```
  --json, -j       Output list as JSON literal
  --installed, -i  Show only installed libraries
  --match, -m      Filter by regular expression (case-insensitive)
  --libraries, -l  List libraries only (no repositories)
  --short, -s      Omit title and description to make list more compact
  --noheaders, -H  Omit header and footer
```
The PR also adds support for hiding repositories from being listed in `qx contrib list`, by adding `(unlisted)` or `(deprecated)` to the repository description on GitHub. I tried to do that with GitHub Topics, but they are not supported by the NPM GitHub API client that we currently use.

Also, it adds the `--version`/`-v` option to `qx` so that now you can do `echo "qx is version $(qx --version)!"` in your scripts.

### DIFF
--- a/docs/cli/contrib.md
+++ b/docs/cli/contrib.md
@@ -88,7 +88,18 @@ command needs to be executed inside your project folder, since it expects
 contrib list` retrieves the version of the qooxdoo framework that the current
 project depends on and filters the list of available contributions accordingly.
 The list displays the names of the repositories and the names of the contained
-contrib libraries that will be installed.
+contrib libraries that will be installed. It has the following options:
+
+```
+Options:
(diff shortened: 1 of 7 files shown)


cboulanger:
DO NOT MERGE YET! I just found some strange behavior-

cboulanger:
Sometimes GIT drives me crazy. No idea why the same set of commits are applied twice. @hkollmann this should be ready to merge (please SQUASH-Merge with the drop-down to get rid of all the commits)

PR #358, opened 2019-04-08
fix qooxdoo path in test-deps.js

hkollmann:
move test-deps.js in front
test it in windows too


cboulanger:
Merged, thank you, @hkollmann!

PR #360, opened 2019-04-08
Add outputFile option to lint command

rad-pat:


lib/qx/tool/cli/commands/Lint.js
```diff
@@ -105,7 +113,11 @@ qx.Class.define("qx.tool.cli.commands.Lint", {
           let outputFormat = this.argv.format || "codeframe";
           const formatter = linter.getFormatter(outputFormat);
           const s = formatter(report.results);
-          if (report.errorCount > 0 || this.argv.warnAsError) {
+          if (this.argv.outputFile) {
+            console.info(this.argv.outputFile);
```
cboulanger: Is this `console.info()` left from debugging or intentional to inform the user? If the latter, please add a `-v`/`--verbose` CLI option (you can copy/paste from other commands) and wrap the output into `if (this.argv.verbose` {}`.

cboulanger: Is this `console.info()` left from debugging or intentional to inform the user? If the latter, please add a `-v`/`--verbose` CLI option (you can copy/paste from other commands) and wrap the output into `if (this.argv.verbose) { console.info("Outputting file into " + this.argv.outputFile)}` or something like that.

hkollmann:
General question: Why do not use piping?

rad-pat:
Honestly, I just pulled the writeFileAsync from elsewhere in the code. I didn't really think too much about it and I don't use node normally so I'm unfamiliar with what's available. Does it make much difference?

lib/qx/tool/cli/commands/Lint.js
rad-pat: This was left over from debugging, but no harm leaving it in if ran verbose so I'll add that option too.

lib/qx/tool/cli/commands/Lint.js
```diff
@@ -114,9 +119,14 @@ qx.Class.define("qx.tool.cli.commands.Lint", {
           const formatter = linter.getFormatter(outputFormat);
           const s = formatter(report.results);
           if (this.argv.outputFile) {
-            console.info(this.argv.outputFile);
+            if (this.argv.verbose) {
+              console.info(`Report to be written to $(this.argv.outputFile)`);
+            }
             await fs.writeFileAsync(this.argv.outputFile, s, "UTF-8")
               .catch(e => console.error(`Error writing report to ${this.argv.outputFile}:` + e.message));
+            if (this.argv.verbose) {
```
cboulanger: sorry to be so picky - I'd either use try/catch here or wrap the second console info into `then()`, since it will be printed even in the case of a write error.

PR #361, opened 2019-04-09
just a little fix if an invalid semver is found

hkollmann:


cboulanger:
Merging, thank you @hkollmann

Issue #353, opened 2019-04-05
Make compiler aware of libraries that are dependencies of dependencies


cboulanger:
This is related to #344 

Dependencies of contribs are downloaded, but not recognized by the compiler. If we don't want the dependencies of contribs (and their dependencies etc.) pollute contrib.json, we need to make the compiler aware of those.

hkollmann:
The compiler is aware of libraries that are dependencies of dependencies. See Compile.js, __checkDependencies

cboulanger:
We need a good test case. I'll create two test contribs that we can use for this

cboulanger:
Ok, maybe it is a good idea to proceed in small steps and agree on what behaviour we want to have, @hkollmann & @johnspackman ?

I have created repos qxl.test1 (with a single library qxl.test1) and qxl.test2 (with 4 libraries qxl.test2A, ...B, ...C, ...D and a qooxdoo.json). As per `Manifest.json/requires`, qxl.test1 has a dependency on qxl.test2/test2C, and qxl.test2A depends on ...B, which depends on ...C, which depends on ...D. 

The libraries have not yet been published to the contrib registry.

No `contrib.json` has been created yet. Since this file is created when you do `qx contrib install`, this is an artificial situation, but one that we should deal with. 

First step:

If I check out the repo `qxl.test1` and to `qx compile`, I get 
```
Error: TypeError: Cannot read property 'qooxdoo/qxl.test2C' of undefined
    at libs.find.entry (/Users/cboulanger/Code/qooxdoo-compiler/lib/qx/tool/cli/commands/Compile.js:637:94)
    at Array.find (native)
    at libs.forEach.lib (/Users/cboulanger/Code/qooxdoo-compiler/lib/qx/tool/cli/commands/Compile.js:637:30)
    at Array.forEach (<anonymous>)
    at defaultConstructor.__checkDependencies (/Users/cboulanger/Code/qooxdoo-compiler/lib/qx/tool/cli/commands/Compile.js:603:12)
    at <anonymous>
```
The error is completely predictable (since the setup is wrong), but I think we should catch it better.

cboulanger:
Ok, this is taken care of by https://github.com/qooxdoo/qooxdoo-compiler/commit/253a41d05ff841dff778fd93c5c3f61f60041a37. Next:
I added source code dependencies between  qxl.test1 -> qxl.test2C and
[...]

PR #362, opened 2019-04-09
Fixes #344 and #353

### LINKED ISSUE #344: dependency infection
Installing eventrecorder contrib v0.9.0 causes all the dependencies of the eventrecorder contrib to be listed in the requirements section of the applications own Manifest.json ... that seems odd as the dependencies of eventrecorder should be its own and not be transmitted up to my application.

cboulanger:
This should make the compiler do what we discussed in #353: Install dependencies of 
dependencies in `contrib.json`, but not in `Manifest.json`, and make it possible to 
delete `contrib.json` and recreate it from the dependency information in `Manifest.json`.

PLEASE TEST THIS BRANCH/PR WITH YOUR APPS FIRST before merging, please.

### DIFF
--- a/lib/qx/tool/cli/commands/Compile.js
+++ b/lib/qx/tool/cli/commands/Compile.js
@@ -618,6 +618,13 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
             requires["qooxdoo-sdk"] = range;
           }
         }
+        // default to empty contrib map
+        if (!qx.lang.Type.isObject(contribs)) {
+          if (this.argv.verbose) {
+            console.warn("No contrib information available.");
+          }
+          contribs={};
+        }
         if (requires) {
           for (let req of Object.getOwnPropertyNames(requires)) {
             let libVer = requires[req];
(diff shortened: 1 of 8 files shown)


cboulanger:
What this doesn't do is to solve version conflicts, i.e if two dependencies require the same library, but with a different version, the version in the `Manifest.json` of the last parent library installed is used and overwrites any other version in `contrib.json`. This has to be solved more intelligently.

cboulanger:
I added some tests (bash only) which not only run the install commands but also check the output.

cboulanger:
@johnspackman @hkollmann @oetiker With the added test I am pretty sure that the PR doesn't break anything, but since the PR involved quite a bit of refactoring and code changes, it would be great if you could build your apps with  the `cboulanger-check-dependency-management` branch once to make sure everything works as before...

hkollmann:
I have read the changes and my projects are still compiling. So let's try it.

cboulanger:
@oetiker Since you had an interest in this issue, can you review, approve and merge? Thanks!

cboulanger:
... the eventrecorder project applications all fine with this PR

PR #364, opened 2019-04-11
Config file versioning

cboulanger:
This is necessary because the last commit introduced changes to the
path of contribs in contrib.json. This PR adds a more general framework
for the future versioning of all configuration files and alerts the user
that manual intervention might be neccessary.

### DIFF
--- a/lib/qx/tool/cli/ConfigSchemas.js
+++ b/lib/qx/tool/cli/ConfigSchemas.js
@@ -1,0 +1,42 @@
+/* ************************************************************************
+
+   qooxdoo - the new era of web development
+
+   http://qooxdoo.org
+
+   Copyright:
+     2019 The qooxdoo developers
+
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+     * Christian Boulanger (info@bibliograph.org, @cboulanger)
+
+************************************************************************ */
+
+/**
+ * This class contains constants used for locating and versioning
+ * of the different configuration files
+ */
+qx.Class.define("qx.tool.cli.ConfigSchemas", {
+
+  statics: {
+
+    library: {
+      filename: "Manifest.json",
+      schema: "https://qooxdoo.org/schema/Manifest.json/v2"
+    },
+
+    compiler: {
+      filename: "compile.json",
+      schema: "https://qooxdoo.org/schema/compile.json/v1"
+    },
+
+    lockfile: {
+      filename: "contrib.json",
+      version: "2"
+    }
+  }
+});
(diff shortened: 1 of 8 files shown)


lib/qx/tool/cli/ConfigSchemas.js
```diff
@@ -0,0 +1,42 @@
...
+ * This class contains constants used for locating and versioning
+ * of the different configuration files
+ */
+qx.Class.define("qx.tool.cli.ConfigSchemas", {
+
+  statics: {
+
+    library: {
+      filename: "Manifest.json",
+      schema: "https://qooxdoo.org/schema/Manifest.json/v2"
```
oetiker: the schema path looks a bit odd ... I would reather expect something like

https://qooxdoo.org/schema/Manifest-json-schema-v2.0.0.json

cboulanger: no, it doesn't. it's just a placeholder. so whenever we/I get to actually create one, the URI will be replaced to point to where it placed (probably on GitHub).

cboulanger: I created the ConfigSchema class to provide a central location for this kind of information, so that it doesn't clutter the code and creates a maintenance headache.

cboulanger:
With the last commit, I made it less scary to upgrade: the script will reinstall the existing libraries into `contrib.json`. This will still erase manual modifications, but at this point, few people will have made any. I think if all tests pass, this PR is ready to be approved & merged.

cboulanger:
I just fixed a  "missing await" bug (https://github.com/qooxdoo/qooxdoo-compiler/pull/364/commits/f514e51eb58e9224600661d50ebee7f99cf1ec37#diff-b2d542a80ea7200f79795275e697b57eR79) with prevented the `qx compile --force` run from working (you always needed a second compile). Now it works! 

1. First time after the new compiler version is used: Warning "The schema of 'compile.json' has changed. Execute 'qx clean && qx compile --force' to delete and regenerate it."
2. `qx clean && qx compile --force` deletes contrib.json but reinstalls all libraries in the versions that were listed in contrib.json (but only saves them to contrib.json, not to Manifest.json). 
3. A backup is made which can be used to reconstruct manual modifications to contrib.json that were not copied over to the new version. 

This should be an ok solution with minimal intervention from the user.

cboulanger:
⚠️ ⚠️  The tests pass but that seems to be wrong: https://travis-ci.org/qooxdoo/qooxdoo-compiler/builds/519133826#L541
Will report when this is fixed.

cboulanger:
Here we are.

lib/qx/tool/cli/commands/Compile.js
```diff
@@ -165,6 +165,11 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
             describe: "Whether bundling is enabled",
             type: "boolean",
             default: true
+          },
+          "force": {
+            describe: "Override warnings",
```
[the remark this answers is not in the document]
cboulanger: Does it really make sense to create a `force*` flag for whatever specific warning we want to override? I'd rather use the know Linux-y way to have a generic  "force" CLI option that can be used in cases where the action is aborted for a specific reason. It will be rare that two separate warning will have to be output simultaneously and then will have to be overridden separately. I think the current naming is fine if we agree to use the generic `--force` only in "fatal" warnings that abort an action.

PR #365, opened 2019-04-13
WIP json schema validation

oetiker:
This PR adds schemas for Manifest.json and compile.json. There are two versions of the schemas.

* Schema version 0 is compatible with the current configs 'out there'.
* Schema version 1 adds a mandatory $schema keyword and
  * removes the deprecated `qooxdoo-version` and `qooxdoo-range` keys from `Manifest.json`
  * adds handles for popular internet sites to the author property
* for version 0 only warnings get issued
* for version 1 and later exceptions are thrown

`compile.json` seems to get loaded by a different mechanism than `Manifest.json` so it does not yet get checked.


resource/Manifest-1.json
```diff
@@ -77,9 +74,9 @@
             }
           }
         },
-        "qooxdoo-versions": {
-          "type": "array",
-          "description": "List of qooxdoo versions this library works with. Deprecated! User requires instead."
+       "version": {
+          "type": "string",
+          "description": "ideally a semver compatible version string"
```
cboulanger: I think we should be strict here : "a semver compatible string" (not "ideally"), so that nobody thinks "trunk" or "master" is an acceptable version

resource/Manifest-1.json
```diff
@@ -1,18 +1,19 @@
 {
   "title": "Manifest.json Schema",
   "$schema": "http://json-schema.org/draft-07/schema",
-  "$id": "https://raw.githubusercontent.com/qooxdoo/qooxdoo-compiler/master/resource/Manifest-1.0.0.json",
+  "$id": "https://raw.githubusercontent.com/qooxdoo/qooxdoo-compiler/master/resource/Manifest-1.json",
   "description": "qooxdoo's Manifest files serve to provide meta information for a library in a structured way. Their syntax is in JSON. They have a more informational part (keyed info), which is more interesting for human readers, a technical part (named provides) that is used in the processing of generator configurations, and a part named externalResources to include CSS and Javascript files.",
```
cboulanger: Why not go with a full version string?

resource/Manifest-1.json
cboulanger: and provide a pattern to validate it: https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39

oetiker: added

resource/Manifest-1.json
oetiker: because after sleeping over it I found that config files are more like database schemas and there it is customary to work with just an increasing number with potentially a system for upgrading the databases between releases ... but I am not deadset against semver here, just thought it might be simpler like this.

cboulanger: You have a point. In the end, it is probably a question of whether the major (non-BC) and minor (BC) version change makes sense in this context. Will it trigger different things? For example, in contrib.json, it seems to make sense since there can be upgrades in information that are still BC (you can use an older file with a newer compiler) or not (you need to regenerate the file).

lib/qx/tool/compiler/utils/Json.js
```diff
@@ -41,15 +43,68 @@ qx.Class.define("qx.tool.compiler.utils.Json", {
       if (!str) {
         return null;
       }
+      // eslint-disable-next-line no-path-concat
+      var qxRoot = __dirname + "/../../../../../";
```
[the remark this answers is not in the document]
oetiker: why would you remove it?

lib/qx/tool/compiler/utils/Json.js
```diff
@@ -41,15 +43,68 @@ qx.Class.define("qx.tool.compiler.utils.Json", {
...
+      var id = this.identifyData(json);
+      if (id) {
+        var ajv = new Ajv({
+          allErrors: true,
+          jsonPointers: true
+        });
+        var schema = "https://raw.githubusercontent.com/qooxdoo/qooxdoo-compiler/master/resource/" + id.type + "-" + String(id.version) + ".json";
+        var valid = ajv
+          .addSchema(require(qxRoot + "resource/compile-" + String(id.version) + ".json"))
+          .addSchema(require(qxRoot + "resource/Manifest-" + String(id.version) + ".json"))
```
derrell: `String(...)` is not required, as numbers are converted to strings when concatenated to other strings.

lib/qx/tool/compiler/utils/Json.js
```diff
@@ -41,15 +43,68 @@ qx.Class.define("qx.tool.compiler.utils.Json", {
...
+      if (data.$schema) {
+        let match = data.$schema.match(/([^-/\s]+)-(\d+)\.json$/);
+        if (match) {
+          return {
+            type: match[1],
+            version: parseInt(match[2])
+          };
+        }
+      }
+      if (data.applications) {
```
derrell: I'm a bit concerned about using information in the file to identify the file type. I'd feel better if the file type were passed in. Consider, for example, an otherwise completely valid compile.json file where the user just forgot to add the `applications` section. It wouldn't handle it as a Manifest file unless it had `provides` in it (but handling it as a Manifest file would show all sorts of errors since it isn't a Manifest file). It would return null, i.e., no known type, and you wouldn't be told the one thing that was wrong with your compile.json file.

oetiker: yes ... I was about to revamp the whole api there ... and require specification of the filetype in the reader call

cboulanger:
Taking over from @oetiker in https://github.com/qooxdoo/qooxdoo-compiler/tree/cboulanger-new-config-system. Will resubmit PR soon!

Issue #363, opened 2019-04-10
Add qx install from external URL/file


cboulanger:
Add the following ways to install a library:
- `qx contrib install https://www.domain.org/path/to/mylibrary.zip`
- `qx contrib install file:///home/user/mylibrary.zip`

This will allow to use libraries from any location. 
See also https://github.com/qooxdoo/qooxdoo-compiler/issues/279#issuecomment-455765841

oetiker:
so while you are at it ... how about making file://home/user/mylibrary work if it is just a directory

cboulanger:
https://github.com/qooxdoo/qooxdoo-compiler/pull/366 does it this way:
`qx contrib install <uri> --from-path path/to/library` allows to use a locally stored version of the contrib
`qx contrib install --from-path path/to/library` simply adds an "anonymous" (in terms of the registry) library. This could later be aliased with `qx contrib install file:///path/to/library`, but that's just syntactic sugar. 
I am not implementing URL support at this point, since there doesn't seem to be an urgent need for it.

cboulanger:
Partially implemented with #366 . https:// installs can go into separate issue, if there is demand.

PR #366, opened 2019-04-15
Partial implementes #363: support install from local path

cboulanger:


johnspackman:

PR #367, opened 2019-04-15
fix error if path do not exist in project

hkollmann:


johnspackman:
merging, thank you @hkollmann

PR #370, opened 2019-04-18
create css dir in resource if dir do not exists.

hkollmann:
This fixes a Problem if we have a scss path but no the target css dir


cboulanger:
Merging bug fix, assuming approval by @johnspackman

cboulanger:
Thank you @hkollmann!

PR #373, opened 2019-04-23
READY FOR REVIEW: Move npm to scoped name

hkollmann:
This commit changes the nm pacakge name of the compiler and use the @qooxdoo/framework package.
Before this can be commited this Change in qooxdoo must be merged https://github.com/qooxdoo/qooxdoo/pull/9665


package.json
```diff
@@ -37,10 +37,11 @@
     "@babel/generator": "^7.4.0",
     "@babel/plugin-transform-block-scoping": "^7.4.0",
     "@babel/polyfill": "^7.4.0",
+    "@babel/preset-env": "^7.4.0",
     "@babel/traverse": "^7.4.0",
     "@babel/types": "^7.4.0",
-    "@babel/preset-env": "^7.4.0",
     "@octokit/rest": "^15.18.0",
+    "@qooxdoo/framework": "file:../qooxdoo",
```
cboulanger: Is that what you want? Looks like you're using a git checkout here....

lib/qx/tool/cli/commands/contrib/Install.js
```diff
@@ -37,7 +37,11 @@ qx.Class.define("qx.tool.cli.commands.contrib.Install", {
     getYargsCommand: function() {
       return {
         command: "install [contrib_uri[@release_tag]]",
-        describe: "installs the latest compatible release of a contrib library (as per Manifest.json). Use \"-r <release tag>\" or @<release tag> to install a particular release.",
+        describe: `installs the latest compatible release of a contrib library (as per Manifest.json). Use "-r <release tag>" or @<release tag> to install a particular release.
+        examples: 
+           * qx contrib install name: Install latest version
```
cboulanger: maybe a bit clearer:
```
                   * qx contrib install name: Install latest published version
	           * qx contrib install name@v0.0.2: Install version 0.0.2,
	           * qx contrib install name@master: Install current master branch from github`,
```

cboulanger:
Also, see https://github.com/qooxdoo/qooxdoo/pull/9665#issuecomment-485741923 which should have been published here.

package.json
hkollmann: You are right - it's the test code. Must be changed before merge. First we need the qooxxdo change.

lib/qx/tool/compiler/app/Application.js
```diff
@@ -779,7 +779,7 @@ qx.Class.define("qx.tool.compiler.app.Application", {
     _applyType: function(value, oldValue) {
       var loader = path.join(__dirname, "loader-" + (this.isBrowserApp() ? "browser" : "server") + ".tmpl.js");
       this.setLoaderTemplate(loader);
-      this.setTheme(null);
+      this.setTheme("");
```
johnspackman: why is this "" instead of null?

johnspackman:

PR #375, opened 2019-04-26
Add --releases-only and --dryrun

cboulanger:
This adds an option to upgrade only the libraries from a released package, 
and leave all others alone. Also, a new dryrun option allows to simulate
an upgrade and see what it would entail. This could be improved by adding
a dryrun option to Install.js


johnspackman:

Issue #369, opened 2019-04-15
Rename "contrib" to something else


cboulanger:

cboulanger:
This is taken from #279 since it addresses a separate issue.

The term "contrib" historically referred to user-contributed libraries. Since we now ship central functionality of the framework, such as the apiviewer as "contribs" (with more to come), it no longer makes sense to talk about "contribs", a weird name anyways that needs explaining. We should be using a more self-explanatory name. 

"contribs" currently refer to 
 - libraries (or collection of libraries) which act as "plugins" or "modules"
 - which are, for the most part, but not always (such as local or remote libraries accessible by URL) registered in a central database so they can easily be installed

we need a term X that would be able to express:
 - a "X" can contain one or more libraries.
 - via the CLI, you install the "X" (with the contained library / all contained libraries) using `qx X install owner/repo`, or a "sub-X" (which would be just one library) via `qx X install owner/repo/path`

Current candidates:
 - "module"
 - "extension"

oetiker:
* package

hkollmann:
package

hkollmann:
With a shortcut to pack or p.

oetiker:
pkg ?

cboulanger:
@johnspackman @derrell @level420 What do you think?

johnspackman:
package & pkg would be fine by me

level420:
I‘m also in favor of package. 
We then could have library packages, theme packages, widget packages etc. 

To be honest: I‘m currently still using generator.py and manually downloaded and manually unpacked contribs which I‘ve referenced manually in the projects config.json. Never used auto-magic things for that.

So I currently trust those working hard to bring the compiler forward. Thank you for that!

derrell:
I'm good with package and pkg. Far more appropriate names, IMO, than contrib.

cboulanger:
Very well. `package`/`pkg` it is, then. Hope to be able to work on it this weekend.

@level420 Once libraries start becoming more modular (See already the qooxdoo demos) and thus depending on each other more heavily, the new functionality will be indispensable. You probably wouldn't want to to be managing 10+ libraries and their dependencies  manually... :-) This will allow to publish small interoperable building blocks of qooxdoo code which will be a huge improvement for the community.

cboulanger:
Implemented by #380

PR #380, opened 2019-04-28
READY FOR REVIEW: Rename contrib to packages

cboulanger:
Implements #369.

This removes all references to "contrib" from the compiler code and the documentation, apart from deprecation notices and the URL of the `qooxdoo/qx-contrib` repository. 

The implementation automatically migrates a few files, making the code backwards-incompatible, i.e.
the code won't work with an earlier compiler version:
- the `contrib` dir is renamed to `packages`
- `contrib.json` is renamed to `qx-lock.json`
- `contrib-cache.json` is renamed to `packages-cache.json`

On first use of a `qx package|contrib` command, the CLI checks if a package needs to be migrated. If yes, `qx package migrate` is called, which implicitly calls `qx package upgrade --reinstall`, which fixes the paths in the lockfile. The migration command 
- automatically corrects the entry for the download cache dir in `.gitignore`,
- fixes the paths in the lock file, and
- updates the dependencies in Manifest.json/requires


hkollmann:
Is it possible to use pkg as an alias for package?

cboulanger:
@hkollmann will fix typo and put in an alias. 

@hkollmann @johnspackman Should I name the download cache dir `qx-packages` instead of `packages`? On second thought, `packages` seems a bit too generic as name...

cboulanger:
Don't merge yet, I have another commit in the works.

cboulanger:
The last commits were a bit chaotic because of a merge conflict with the shrink-wrap file. 

Now we have a migration that works async (which requires a bit of heavy refactoring) and will also update the `.gitignore` file (could do others as well).

cboulanger:
Last commit of the day: now the migration automatically fixes the paths in the lock file. What we still need is an automatic update of the required version of the compiler, since the project state after migration is incompatible with earlier versions.

cboulanger:
@johnspackman @hkollmann Given that this PR is backwards-compatible to existing code, but produces backwards-incompatible user code, I now think I should rebase it on @hkollmann’s PR and apply the changes to the new scoped NPM packages only. What do you think?

johnspackman:
@cboulanger re rebasing: that could be a good plan, so that the old `qxcompiler` package doesn't have ay major changes immediately before becoming deprecated.  Which means it would be good to create a branch from the last qxcompiler release with a final change that warns that it is deprecated and the name change

johnspackman:

johnspackman:
I just realised that "package" is a reserved word in ES5, does this make it illegal to have `qx.tool.cli.commands.package`?  Perhaps rename to `qx.tool.cli.commands.packages`?

cboulanger:
It's not a problem for object properties: https://stackoverflow.com/a/17911822

PR #384, opened 2019-04-30
Compiler fixes 2

johnspackman:


lib/qx/tool/cli/commands/Compile.js
```diff
@@ -70,6 +66,11 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
             type: "string",
             array: true
           },
+          "update-po-files": {
+            describe: "enables detection of translations and writing them out into .po files",
```
cboulanger: Maybe provide an alias such as `-u` since this is a long option name?

johnspackman: no problemo

cboulanger:
Could you maybe add aliases to the compiler commands, at least for the ones that are likely used most?

PR #390, opened 2019-05-08
add createElement stub

hkollmann:
To use @qooxdoo/framework in qxcompiler we need this patch as a first step. @qooxdoo/Framework is not running if not compiled with a createElement stub


cboulanger:
Merged, thank you @hkollmann

Issue #394, opened 2019-05-12
Implement new configuration system including JSON-Schema validation


cboulanger:
Implement a better configuration system that avoids constant re-reading and re-writing of the config files by content caching and implicitly checks the well-formedness of the files using JSON-Schema.

Supersedes #365 and #190

hkollmann:
Done with #416

PR #395, opened 2019-05-12
move version to 1.0.0-beta

hkollmann:


cboulanger:
@hkollmann can you replace Update.js, line 319 with
```javascript
              if (argv.verbose && data.requires && data.requires["qooxdoo-sdk"]) {
```

cboulanger:
In the implementation #394, I'll make sure to have a look at all checks to prevent this sort of crash if the config file is malformed.

cboulanger:
Merged after disabling a non-functional test, thank you @hkollmann !

PR #397, opened 2019-05-15
Add support for class files ending in .ts

thraaawn:
To make use of TypeScript in an existing qooxdoo project, especially
when migrating incrementally, it would be useful to allow .ts files
to be transpiled by the qooxdoo compiler.

This does not require much voodoo since babel 7 has presets that allow
stripping of typescript type annotations et al. More details can be
found for example in the announcement post here:
https://devblogs.microsoft.com/typescript/typescript-and-babel-7/

What needs to change then is:
- the compiler being able to recognize .ts files in the same way as
  .js files and reconstruct the filename including extension
- the babel step needs to use the @babel/preset-typescript preset

This allows users to rename their .js files to .ts and make use of the
languages features, especially type annotations. An IDE like Visual
Studio Code will properly type check those files. The typescript
compiler can be used separately to type check the .ts files during
build steps.

### DIFF
--- a/lib/qx/tool/compiler/ClassFile.js
+++ b/lib/qx/tool/compiler/ClassFile.js
@@ -291,7 +291,8 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
               filename: t.getSourcePath(),
               sourceMaps: true,
               "presets": [
-                [ require.resolve("@babel/preset-env"), options]
+                [ require.resolve("@babel/preset-env"), options],
+                [ require.resolve("@babel/preset-typescript")]
               ],
               plugins: [
                 t._babelClassPlugin()
@@ -2015,7 +2016,7 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
      * @returns {String}
      */
     getSourcePath: function(library, className) {
-      return pathModule.join(library.getRootDir(), library.getSourcePath(), className.replace(/\./g, pathModule.sep) + ".js");
+      return pathModule.join(library.getRootDir(), library.getSourcePath(), className.replace(/\./g, pathModule.sep) + library.getSourceFileExtension(className));
     },
 
     /**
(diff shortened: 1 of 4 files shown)


johnspackman:
I like this, there's only one thing I would do differently and that's to not add a requirement for a library to be 100% TS vs 100% JS; IMHO it should be able to detect which extension to use automatically.

Looking at the code, the `getSourcePath()` *static* method is only used in the constructor to populate `this.__sourceFilename` and that in turn is not needed until the `.load` method is called.  This means that you could delete the *static* `getSourcePath()` and then use `.load` to detect which to use.

lib/qx/tool/compiler/ClassFile.js
```diff
@@ -2015,7 +2016,7 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
      * @returns {String}
      */
     getSourcePath: function(library, className) {
-      return pathModule.join(library.getRootDir(), library.getSourcePath(), className.replace(/\./g, pathModule.sep) + ".js");
```
johnspackman: as commented, delete this and move to `.load()` with auto detection

lib/qx/tool/compiler/app/Library.js
```diff
@@ -342,6 +345,16 @@ qx.Class.define("qx.tool.compiler.app.Library", {
...
+    /**
+     * Returns the original extension of the class file that implemented the
+     * given class name.
+     *
+     * @param {String} className
+     */
+    getSourceFileExtension: function(className) {
+      return this.__sourceFileExtensions[className];
+    },
+
```
johnspackman: as commented, this would not be necessary with auto detection

thraaawn:
Hey, thanks for the feedback.

I did not want to change that much about the code, i.e. did not want to change existing methods and contexts. I'm not sure where getSourcePath() is being used in the first place and would need to look into your suggestion for a while to understand where, when and why it is used.

I'm not sure I properly understand: This currently does not force a library to be 100% TS or 100% JS, it remembers the original extension of the class file so it can reconstruct it later.

Not sure what you mean by auto detect - do you mean analysing the class itself? Wouldn't that bring significant performance implications instead of looking at the original files extension?

johnspackman:
> I'm not sure I properly understand: This currently does not force a library to be 100% TS or 100% JS, it remembers the original extension of the class file so it can reconstruct it later.

Ah, sorry I misread the code.

PR #398, opened 2019-05-15
move eslint modules to new @qooxdoo namespace

hkollmann:
I moved the eslint qooxdoo packages into the @qooxdoo Namespace. Now the compiler should use this


cboulanger:
Full approval, merged, thank you @hkollmann!

PR #402, opened 2019-05-17
Fix js compile

johnspackman:


hkollmann:
Could you try to compile the https://github.com/qooxdoo/qxl.apiviewer with this? It uses compile,json excessive.

johnspackman:
done; there were some changes needed, I've pushed a fix for the API viewer

hkollmann:
Then the demobrowser https://github.com/qooxdoo/qxl.demobrowser will not compile and longer either

Von meinem iPhone gesendet

hkollmann:
When we have inputdata in the global scope we can remove it from the call to the function. 
This change breaks all compile.js - so we can remove Double data 

Von meinem iPhone gesendet

johnspackman:
sorry, i broke that rebase ...

PR #403, opened 2019-05-17
Typescript: Keep method documentation and add getInstance method to singletons

thraaawn:
I'm currently on a quest to migrate some of our code-base to typescript and there are some things blocking a smooth transition. I'm not sure how your contribution guidelines for these kind of things are, so I'm just PRing these as I go. There will probably be more PRs of this type in the future, if that is in your interested.

This particular PR takes care of classes of type singleton missing the type definition for the getInstance method. This brought a small bug to light (regarding isSingleton on class meta data being wrong) that seemingly never caused any issues. That is fixed and is being used by the TypeScriptWriter now.

The other commit in this PR ensures that the method documentation isn't lost when generating the type definition file.


johnspackman:
2 approvals, merging thank you @thraaawn !

PR #404, opened 2019-05-17
Groundwork for new config system

cboulanger:
This lays the groundwork for the new config file system based on JSON-Schema validation. The necessary infrastructure is there now, but I haven't yet replaced the old machinery yet. We need a squash merge first to update master now and then I will work on the rest. 

The JSON schema files are located in `resource/schema/<version>/` => `https://raw.githubusercontent.com/qooxdoo/qooxdoo-compiler/master/resource/schema/<version>/Manifest.json`, `<version>` referring to the schema version (currently, "v1"). It should eventually be mirrored to `https://qooxdoo.org/schema/<version>/Manifest.json` which will result in a much nicer "$schema" string. 

There is a problem with backslashes in regular expressions in the JSON schema files (regexes with backslashes in them just wont match anything), which I need to investigate, but I have fixed them where they were relevant for the time being.


johnspackman:

lib/qx/tool/compiler/utils/Json.js
```diff
@@ -17,69 +17,152 @@
...
+
+    /**
+     * Identify the type and version of the config file schema in the data that
+     * has been passed. Return an object containing information on that schema
+     * which can be used to validate the data.
+     * @param data {Object} Manifest.json or compile.json
+     * @return {{type,version,file,url}|null} Object with information about the
+     * schema or null if no schema could be detected.
+     */
+    getSchemaInfo: function(data) {
```
oetiker: is that still needed?

cboulanger: yes, this method provides metadata on the config file, such as type, version, the path to the local copy of the schema file of the schema., I realize that computing the `url` return value in line 131 doesn't seem to  make much sense, since it is in the `$schema` property already. I have to think about whether is should be discarded or if it is a way to check the `$schema` value.

oetiker: I was more wondering about the auto-sensing capability ... I guess that should not be needed anymore as the one reading the config file knows which one is wanted.

cboulanger: You are right - there is a logic error here - ATM, we get the schema from the actual data which is loaded into the model and check afterwards if it is the right one whereas the config file model knows which schema it should and should enforce it right away. I'll correct it accordingly.

PR #406, opened 2019-05-20
Fix js compile 2

johnspackman:
Replaces https://github.com/qooxdoo/qooxdoo-compiler/pull/402

### DIFF
--- a/docs/configuration/compile.md
+++ b/docs/configuration/compile.md
@@ -379,22 +379,142 @@ If you choose to use the optional web server by running `qx serve`, you can chan
 ```
 
 ## compile.js
-Configuration files do not support processes, job executions, or even macros - if you want to add basic processing (eg for macros), 
-use a .js file to manipulate the data.  When your `compile.js` is executed, there will be a global object called `compiler` which
-has these properties:
+Unlike the python generator's config.json, compile.json does not support processes, job executions, or even macros.  If you want to add 
+basic processing (eg for macros), you can use a `compile.js` file to manipulate the data in `compile.json`, or even replace `compile.json`
+altogether.  
+
+If you provide a `compile.js`, the compiler will first load the `compile.json` (if there is one) and then run the code in `compile.js`; when
+`compile.js` is executed, the compiler will pass the contents of `compile.json` so that it can be manipulated and returned (of course, you
+could chose to construct a brand new object instead). 
+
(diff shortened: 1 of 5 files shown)


cboulanger:
@johnspackman Is there a reason why you don't `require()` compile.js like a node module, which exports a single function? (This is not a change request, just to understand)

docs/configuration/compile.md
```diff
@@ -379,23 +379,88 @@ If you choose to use the optional web server by running `qx serve`, you can chan
...
+        
+    } else {
+        return { /* ... snip ... */ };
+    }
+)();
+```
+
+Or an example using callbacks:
+```
+function(inputData, cb) {
```
hkollmann: inputData should be removed here too,.

johnspackman:
@cboulanger originally my goal was to produce usable error messages (eval is useless in that regard), but using the `vm.Script` also allows the `compile.js` to be run in an isolated context, where only those variables you want to make available to the script are available - this includes global variables.

In [qxl.apiviewer/compile.js](https://github.com/qooxdoo/qxl.apiviewer/blob/master/compile.js#L3) this meant that even the global `qx` was not available and `require("@qooxdoo/framework");` had to be added.

I'll take responsibility for fixing demobrowser etc if this PR is accepted - but also I'm happy to switch to `require("./compile.js")` if that's what everybody else would prefer.  Mostly I wanted compile.js to have real error messages and line numbers!

hkollmann:
Personally i prefer using build in Methods like require instead of external libraries. Much less stuff to maintain …

cboulanger:
I am happy with leaving things as they are - but as the solution is a bit unusual (at least for me), I would explain them a bit more detailed as I have tried above - does that make sense to you, @johnspackman ?

johnspackman:
@hkollmann Personally i prefer using build in Methods like require instead of external libraries. Much less stuff to maintain …

It is a built in library, it's the official node way of running external scripts... https://nodejs.org/dist/latest-v10.x/docs/api/vm.html

cboulanger:
@johnspackman I see! Now I understand the mechanism. I thought the code would be eval'd. Maybe you can add to the documentation that this is how `compile.js` is run?

cboulanger:
Updated my suggestion above to reflect what I learned 😄

cboulanger:
2 approvals, merged PR. Thank you @johnspackman !

PR #407, opened 2019-05-21
This change exposes the global `qx` and `process` objects

johnspackman:
the `qx` object is exposed so that the `compile.js` has the exact qooxdoo the compiler uses, and so that it has the `qx.tool.*` classes available to it.  The `process` object is normally available to node apps.


johnspackman:
PLease don't merge this just yet - I've found another issue I need to fix first ...

cboulanger:
Sorry, too late - should I open a revert PR oder we just wait until you fix the issues in a new PR?

johnspackman:
no, it won't hurt, but there is something else to follow...

PR #405, opened 2019-05-19
New directory layout

cboulanger:
This implements the new directory layout as discussed. Tests pass, but it is very likely that there are bugs left from moving files and folders and renaming classes accordingly. Please test this PR thoroughly before approving it by checking out the branch `cboulanger-new-dir-layout`.

I hope everything is nice and tidy now!

@johnspackman I have moved the `tool/compiler/<name of subdir>.js`  index files to `tool/compiler/<name of subdir>/index.js` since this seemed more logical to me in terms of structuring submodules, I hope you are ok with this change. Also, I moved a lot of stuff into the `test` folder (like `demos` and `utils`). Some of it looked like you haven't used it in a while.

PR #409, opened 2019-05-21
Switches to using `require` instead of `require('vm').Script` to load compile.js

johnspackman:
Switches to using `require` instead of `require('vm').Script` to load compile.js; this is because while `vm` allows sandboxing code into a separate context, there is no implementation of `require()` in that sandbox.  The recommended way is to pass `require` in, but that breaks with Qooxdoo classes because they are loaded into the global scope of the compiler, which means that `require()` in compile.js will not update the global scope for compile.js.

EG the code for compile.js in API viewer was failing because this code:
```
require(path.join(lib.getRootDir(), lib.getSourcePath(), "qxl/apiviewer/ClassLoader.js"));
```
would return an empty object and `qxl.apiviewer.ClassLoader === undefined`; this happened because `require()` loads into the compiler scope and not into `compile.js` scope.

As compile.js is now a module, this means changing the implementation of the API and docs

### DIFF
--- a/docs/configuration/compile.md
+++ b/docs/configuration/compile.md
@@ -466,7 +466,7 @@ module.exports = function(compiler) {
 ### How to add sass call for mobile projects:
 
 ```javascript
-async function compile(data, callback) {
+module.exports = async function compile(compiler) {
   /**
  * Adds sass support for current project.
  * Needed for qx.mobile projects.
@@ -475,12 +475,13 @@ async function compile(data, callback) {
  *    - add dependency to project package.json: "runscript": "^1.3.0"
  *    - run npm install in project dir.
  *  
- * @param {*} data       : config data from compile.json
+ * @param {*} data       : compiler interface
  * @param {*} callback   : callback for qxcli.  
  */
   debugger;
   const runscript = require('runscript');
   const util = require('util');
+  let data = compiler.inputData;
   runScript = async function (cmd) {
     return new Promise((resolve) => runscript(cmd, {
         stdio: 'pipe'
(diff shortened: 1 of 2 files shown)


cboulanger:
I assume this is not BC - we need to mention it in the blog post announcing \@qooxdoo/framework v6.0.0-beta and \@qooxdoo/compiler v1.0.0-beta.

johnspackman:
ah - I'd forgotten we're now beta.  You're right that it's not BC, neither was the previous one.  I think it's worth keeping in though, because it makes typos in the compile.js easily detectable

cboulanger:
@johnspackman I think that's ok as we transition from alpha to beta, and we haven't told people to switch to the new scoped packages yet. But from now on we should not change the API unless  it is absolutely necessary.

hkollmann:
Any reason why do you not use script.call(this, inputData) as before? Then compile script runs direct in Compiler context. 
IMHO this easier to understand than the compiler variable.

johnspackman:
i very nearly started with a reverse commit (ie go completely back) but I was thinking that this way it allows for future expansion, and as the api has to change anyway it seemed like something we could do just once.

cboulanger:
Having `this` refer to the compiler in the `require`d function wouldn't be any more intuitive than passing an object to the function, I think , although it would of course create less costly to stay BC.

hkollmann:
Any reason why not use script.call(this, inputData any longer? So compile.js is running in the context of the compiler.
IMHO this is easier to read then the compiler variable.

hkollmann:
I prefer the BC version...

cboulanger:
We could have both, stay BC, but support and document the parameter-passing solution `(data, cb, compiler)` - somewhat unusual to not have the callback as last argument, but that's just a convention.

hkollmann:
Would be a Little bit redundant.

cboulanger:
That's true, but that's often the case when you change things AND try to stay BC. :-)

hkollmann:
let @johnspackman  decide..

cboulanger:
I am fine with both "no change" / "a bit of redundancy". All I care about at the moment is to get #405 merged so that we can move on with other stuff.

Issue #411, opened 2019-05-22
"externalResources" key in Manifest.json


cboulanger:
There seems to be some confusion about the location of the `externalResources` in Manifest.json: is it a property of `provides` or a top-level property? 

The documentation is inconsistent: 
a. In the [example](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/docs/configuration/Manifest.md#reference), it is a top-level property.  All our templates treat it as a top-level property, and `qx add script` as well as the [compiler code](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/app/Library.js#L201) do so, too. 
b. In the [description](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/docs/configuration/Manifest.md#description-of-keys) and in the [json schema](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/resource/qx/tool/schema/v1/Manifest.json) it belongs to `provides`. It might make more sense to put it into the `provides` object, but this would break alot of people's code, unless we provide an automatic migration path.

How should we proceed? I have fixed the json schema in this branch: https://github.com/qooxdoo/qooxdoo-compiler/tree/cboulanger-apply-config-api . We could also allow both and document only b).

hkollmann:
It's a top level property. It do not belong to provides. IHMO should stay where it's implementated.

cboulanger:
Issue has been settled via #413. @johnspackman, could you update the documentation?

johnspackman:
no problem, its on my list

johnspackman:
oh wait a minute - isnt https://github.com/qooxdoo/qooxdoo-compiler/pull/414 it by @hkollmann ??

hkollmann:
Yes it is

Von meinem iPad gesendet

PR #413, opened 2019-05-22
Fix manifest schema

cboulanger:
This fixes the JSON schema of `Manifest.json` in line with #411 - if we agree that `externalResources` is a top-level property.

Also, it fixes some of the regular expressions which did not work using backslash escaping.


cboulanger:
Ups, something else crept in there, too. But it's no harm since it is part of another pending PR.

cboulanger:
More explicitly, the other change is that `qx migrate` is removed, as agreed with @johnspackman

PR #414, opened 2019-05-22
fix documentation of externals

hkollmann:


johnspackman:
merging, thank you @hkollmann

Issue #412, opened 2019-05-22
New qx.tool.config namespace


cboulanger:
I would like to make a final change in the config system. Instead of storing config file metadata in a [central file](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/ConfigSchemas.js) and initialize the config model using an factory function and a type id like so
```
let manifestModel = await qx.tool.utils.ConfigFile.getInstanceByType("manifest",/*do not load?*/ false, /*warn Only*/ false);
```
I'd like to introduce a new namespace `qx.tool.config` and declare a separate class for each config file, which inherits from an abstract class, and initialize it like so:
```
let manifestModel = await (new qx.tool.config.Manifest({warnOnly:false})).load();
```
This would be much more self-documenting and I think much more intuitive. What do you think?

cboulanger:
thinking about it: the models should be singletons so that they can be used from everywhere without having to pass around the object:

Example:

custom.Main.js:
```
let manifestModel = await qx.tool.config.Manifest.getInstance().set({warnOnly:false}).load();
manifestModel.setValue("foo", "bar");
custom.Other.method();
// do other stuff
manifestModel.save();
```
custom.Other.js:
```
let manifestModel = await qx.tool.config.Manifest.getInstance();
manifestModel.setValue("baz", boo");
```

hkollmann:
Done with #416

PR #416, opened 2019-05-22
Config API final change

cboulanger:
Implements #412.


cboulanger:
I have added a test for migration: https://github.com/qooxdoo/qooxdoo-compiler/blob/cboulanger-config-api-final-change/test/test-pkg-migrate.js which uses qxl.apiviewer v0.1.14

PR #417, opened 2019-05-24
adds to `qx package migrate` so that a compile.js without exports is detected

johnspackman:
also fixes a typo in manifest schema

### DIFF
--- a/source/class/qx/tool/cli/commands/package/Migrate.js
+++ b/source/class/qx/tool/cli/commands/package/Migrate.js
@@ -20,12 +20,8 @@ require("../Package");
 require("@qooxdoo/framework");
 const process = require("process");
 const path = require("upath");
-<<<<<<< Updated upstream
 const semver = require("semver");
-const fs = require("fs");
-=======
 const fs = qx.tool.utils.Promisify.fs;
->>>>>>> Stashed changes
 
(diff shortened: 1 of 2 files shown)


source/class/qx/tool/cli/commands/package/Migrate.js
```diff
@@ -180,6 +180,13 @@ qx.Class.define("qx.tool.cli.commands.package.Migrate", {
           }
         }
       }
+      let compileJsFilename = path.join(process.cwd(), "compile.js");
+      if (await fs.existsAsync(compileJsFilename)) {
+        let data = await fs.readFileAsync(compileJsFilename, "utf8");
+        if (data.indexOf("module.exports") < 0) {
+          console.warn("*** Your compile.js appears to be missing a `module.exports` statement - please see https://git.io/fjBqU for more details");
```
cboulanger: You can add `needFix = true;` here then it will not proceed until the user fixes the issue (which s/he should).

cboulanger: wait, no, that's not how the code works.

cboulanger: I'll adapt it later...

PR #418, opened 2019-05-24
some fixes 

hkollmann:
- add $schema to templates
- Abstract.js: needs upath for windows
- Abstract.js: return boolean instead of map if parameter is object
- Migrate.js: fix ( bug

### DIFF
--- a/source/class/qx/tool/config/Abstract.js
+++ b/source/class/qx/tool/config/Abstract.js
@@ -18,7 +18,7 @@
 
 const fs = qx.tool.utils.Promisify.fs;
 const process = require("process");
-const path = require("path");
+const path = require("upath");
 const semver = require("semver");
 const get_value = require("get-value");
 const set_value = require("set-value");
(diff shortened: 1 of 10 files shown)


source/class/qx/tool/config/Abstract.js
```diff
@@ -403,10 +403,12 @@ qx.Class.define("qx.tool.config.Abstract", {
       if (qx.lang.Type.isString(propOrMap)) {
         return this.getValue(propOrMap) !== undefined;
       } else if (qx.lang.Type.isObject(propOrMap)) {
+        res = false;
         for (let key of Object.getOwnPropertyNames(propOrMap)) {
           propOrMap[key] = this.keyExists(key);
+          res = res || propOrMap[key];
         }
-        return propOrMap;
+        return res;
```
cboulanger: Of course, this couldn't work, good catch!

cboulanger:
Merged, thank you, @hkollmann !

PR #419, opened 2019-05-25
Fix dependencies

cboulanger:
This removes the annoying "[before-after-hook]: "Hook()" repurposing warning, use "Hook.Collection()". Read more: https://git.io/upgrade-before-after-hook-to-1.4"  by downgrading the `before-after-hook` to <v1.4.0 and upgrades other dependencies in response to security alerts.


johnspackman:

PR #420, opened 2019-05-25
Fix regressions

cboulanger:
This fixes a couple of regressions that were introduced by applying the new config API. Please merge #419 first.


johnspackman:

PR #422, opened 2019-05-26
Small fixes to prevent migration warnings + formatting

cboulanger:
This will prevent to have to call `qx pkg migrate` and `qx compile --force` to upgrade the two files.


johnspackman:

PR #425, opened 2019-05-28
Documentation update

cboulanger:
This updates the documentation to the current state of the code. Still alot is missing, in particular better documentation on migrating from v5 to v6, and even more specifically, how to transition from `package.json` to `compile.(json|js)`. It would be great if @johnspackman could add something to this effect.

We need to think about how to integrate this documentation with @oetiker's new doc system. Should we maintain it here and periodically copy it over to a dedicated documentation repo? How to synchronize versions and documentation, then? One great thing about the new system is the button "improve documentation"  in the docs themselves - but this only works if the docs are in one repo...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # Qooxdoo Compiler and Command Line Interface
 
-**NOTE** If you have previous installed `qooxdoo-cli`, please uninstall it before continuing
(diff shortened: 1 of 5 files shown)


johnspackman:

PR #426, opened 2019-05-29
run lint --fix and lint test/*.js --fix

hkollmann:
and fix all open errors.

No there are only warnings About todo and fixme left.

We should rework them as a next step.


cboulanger:
Full approval, merging. Thank you  @hkollmann !

Issue #309, opened 2019-02-14
"Unresolved use of symbol" for default argument


cboulanger:
Given this code:
```javascript
qx.Class.define("cboulanger.eventrecorder.UiController", {
  extend: qx.ui.window.Window,
  construct: function(caption="Event Recorder") {
    this.base(arguments);
    // assign id to this widget from caption
    const objectId = caption.replace(/ /g, "").toLocaleLowerCase();
```
the compiler complains "Unresolved use of symbol caption.replace"

hkollmann:
Compiling qx with qx gives another example:

qx.tool.cli.commands.Command: [380,53] Unresolved use of symbol replaceInFilesArr.length

    async migrate(fileList, replaceInFilesArr=[], annouceOnly=false) {
      let quiet = this.argv.quiet;
      if (qx.lang.Type.isArray(fileList)) {
        let filesToRename = this.checkFilesToRename(fileList);
        if (filesToRename.length) {
          if (annouceOnly) {
            // announce migration
            console.warn(`*** Warning: Some metadata filenames have changed. The following files will be renamed:`);
            for (let [newPath, oldPath] of filesToRename) {
              console.warn(`    '${oldPath}' => '${newPath}'.`);
            }
          } else {
            // apply migration
            for (let [newPath, oldPath] of filesToRename) {
              try {
                await fs.renameAsync(oldPath, newPath);
                if (!quiet) {
                  console.info(`Renamed '${oldPath}' to '${newPath}'.`);
                }
              } catch (e) {
                console.error(`Renaming '${oldPath}' to '${newPath}' failed: ${e.message}.`);
                process.exit(1);
              }
            }
          }
        }
      }
      if (qx.lang.Type.isArray(replaceInFilesArr) && replaceInFilesArr.length) {
        for (let replaceInFiles of replaceInFilesArr) {
          if (annouceOnly) {
            console.warn(`*** In the file(s) ${replaceInFiles.files}, '${replaceInFiles.from}' will be changed to '${replaceInFiles.to}'.`);
            return;
          }
          try {
            let results = await replace_in_file(replaceInFiles);
            if (!quiet) {
              let files = results.filter(result => result.hasChanged).map(result => result.file);
              console.info(`The following files were changed: ${files.join(", ")}`);
            }
          } catch (e) {
            console.error(`Error replacing in files: ${e.message}`);
            process.exit(1);
          }
        }
      }
    }
  }
});

[...]

PR #427, opened 2019-05-29
Fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/309

johnspackman:


test/testapp/qx-lock.json
```diff
@@ -1,11 +1,16 @@
...
+      "path": "qx_packages/qooxdoo_qxl_testtapper_v0_3_1",
+      "uri": "qooxdoo/qxl.testtapper",
+      "repo_name": "qooxdoo/qxl.testtapper",
+      "repo_tag": "v0.3.1"
     }
-  ]
-}
+  ],
+  "version": "2.1",
+  "requires": {
```
cboulanger: This is wrong here-should have gone into the manifest....

johnspackman: weird - I didn't edit that file, just ran `qx migrate`

johnspackman:
@cboulanger @hkollmann why has this suddenly failed with lint errors?  Have the lint rules been changed?

hkollmann:
Rules have'nt change. We just check lint now.

cboulanger:
I think we should disable the no-todo rule. I think it makes sense to not allow FIXMEs, but TODOs should be fine.

hkollmann:
no-todo is just a warning. not an error.

cboulanger:
@hkollmann Ah ok, thanks!

johnspackman:
> no-todo is just a warning

yet another example of why lint is 99% useless.  Apparently, having a blank line before `} else if (...)` is  an error too.

> Strange fail in the test, Ill restart.

The fail is because the test is broken by this commit, fixing that now.

cboulanger:
@johnspackman We can disable individual rules if we don't find them useful. In all, I find eslint really helpful to spot problems. If we really need all those strict curly braces rules, I don't know.

PR #429, opened 2019-05-30
add call to lint into test

hkollmann:


johnspackman:

PR #430, opened 2019-05-30
fix layout of babel options

hkollmann:


johnspackman:

PR #432, opened 2019-06-01
Fix the schema urls to contain the schema version in the file name, not in the path

cboulanger:
The schema version is encoded in the file name following https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/


johnspackman:

PR #434, opened 2019-06-05
__external__ is not resolved. 

hkollmann:
This uses the old way and add libnamespace instead of __external__. Same behavior as css.


hkollmann:
In my project it‘s not resolved in Boot.js

Von meinem iPhone gesendet

johnspackman:
can you give me a reproducable example?

hkollmann:
I can send you my project via pm later

Von meinem iPhone gesendet

johnspackman:
@hkollmann thanks for your project pm 👍 

I have a fix which I think fixes it (rolled in with a couple of other minor bits), please see https://github.com/qooxdoo/qooxdoo-compiler/pull/436

hkollmann:
solution in #436 !

PR #436, opened 2019-06-05
Compiler fixes 4

johnspackman:
Please see individual commits for details


source/class/qx/tool/compiler/targets/Target.js
```diff
@@ -395,12 +396,15 @@ module.exports = qx.Class.define("qx.tool.compiler.targets.Target", {
             var arr = library.getAddScript();
```
hkollmann: IMHO this should be done for the Css Array to. They should also be enabled to use externals.

hkollmann:
Could you run qx lint --fix?

Issue #401, opened 2019-05-17
Merge 'new qx.tool.cli.Utils.UserError' and 'qx.tool.compiler.utils.UserError'


cboulanger:
1. `qx.tool.cli.Utils.UserError` returns a constructor that needs to be initialized with `new` 
2. `qx.tool.compiler.utils.UserError` is a factory function that returns an error instance

I would argue to use 1) because people are used to write `throw new CustomError` and not `throw customErrorFactoryFunction`

cboulanger:
Wait, no, that is not correct. Both return an instance, because BOTH are factory functions which instantiate an error object and then remove the stack trace. 

We should replace BOTH factory function with a real class which inherits from `Error` and removes the stack in the constructor.

cboulanger:
Both `throw new qx.tool.cli.Utils.UserError("Oh No!");` and `throw qx.tool.cli.Utils.UserError("Oh No!");` works because in the first case it simply instantiates a new o object from the new object.

hkollmann:
qx.tool.cli.Utils.UserError is gone with new structure. So we only have to deal with qx.tool.utils.Utils.UserError yet

cboulanger:
Fixed with #437.

PR #437, opened 2019-06-06
use a class instead of factory method for user error

hkollmann:


cboulanger:
Since the test pass, this seems to work. However, I'd rather use a separate class as error. A class as a member of a class is a weird construct, I think. This does fix #401, but I'll create a separate issue for a new separate class `qx.tool.utils.UserError`

cboulanger:
2 approvals, merging. Thank you @hkollmann

PR #447, opened 2019-06-14
Compile.js regression

johnspackman:


johnspackman:
Then IMHO API Viewer needs some review - it was always the case that compile.json was loaded and passed to compile.js, so that compile.js is able to augment/process the compile.json.  I have made this change before when I fixed the compile.js, so I guess from your comment that you must have reverted this in a recent PR and I missed that's what had happened.  

The issue here is that compile.js/json act as a pair (so .json is read and passed to .js), and then the result is a configuration structure which can be parsed and normalised, so that it can subsequently be used to configure the compiler API.

But the reason that this was a regression (and why I fixed it, this time and before) is that `compile.js` is being ignored completely, unless it also happens to be in a library.

In the current master, this will lead to unexpected side effects - for example, if you run `qx compile` in your app, then the API viewer's `compile.js` will be run with the application's `compile.json` -- but if you run the same command in the API viewer `compile.js` will be run with the API viewer's .json.  This means that the API viewer's compile.js will be run with completely different (and undefined) source data.

The other reason that the current master is broken is because the `compile.js` needs to be able to modify the structure *before* `_mergeArguments` is called, because `_mergeArguments` normalises the configuration as well as integrating command line arguments - for example, it makes sure that there is a target, etc.  However, libraries are not known until *after* `_mergeArguments` has normalised the configuration.

AFAICT the problem which API viewer's compile.js is solving is that it needs to augment the entry for it's application, by updating the application's environment entry; for the API viewer this needs to be done each time the target fires `"checkEnvironment"`, and it just so happens that API viewer ignores the `inputData` so API viewer is fine with the way that master works.

IMHO in order to allow libraries to contribute to applications (eg API Viewer) and allow processing of compilation, the API needs to be improved so that there are phases:

(1) compile.json has been processed (if it exists) and normalisation of the structure has not yet begun;
(2) normalisation complete, each library's compile.js is asked to contribute

This will mean that if a library has it's own `compile.js` used for building it's own demo/test apps, then running `qx compile` in the library directory will cause both phases to be fired on that library's `compile.js`

 - but crucially, if a compilation takes place outside the library (an application that just uses the library), then the library will only see the second phase.

source/class/qx/tool/cli/commands/MConfig.js
```diff
@@ -112,14 +116,6 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
       }
       this._mergeArguments(parsedArgs, config, lockfile_content);
 
-      if (config.libraries) {
-        for (const aPath of config.libraries) {
-          if (typeof aPath === "object" && typeof aPath.path === "string") {
```
cboulanger: Can we simply remove this warning now - do you think no-one has the old format anymore? Or is it unnecessary since `qx migrate` has to be run anyways and take care of it?

cboulanger:
So is the problem that the apiviewer depends on code in `compile.js` which isn't run when [this line](https://github.com/qooxdoo/qooxdoo-compiler/commit/cec91963e924d51a31f03c546844d779664cc780#diff-25f3dd808436d62e36636886c807ba37L120) is removed? 

This means we need so kind of "hook" for packages that act as "plugins", don't we?

cboulanger:
Something strage here - the PR could be merged even though we don't have two approvals...

hkollmann:
> Then IMHO API Viewer needs some review - it was always the case that compile.json was loaded and passed to compile.js, so that compile.js is able to augment/process the compile.json.  I have made this change before when I fixed the compile.js, so I guess from your comment that you must have reverted this in a recent PR and I missed that's what had happened.
> The issue here is that compile.js/json act as a pair (so .json is read and passed to .js), and then the result is a configuration structure which can be parsed and normalised, so that it can subsequently be used to configure the compiler API.
> But the reason that this was a regression (and why I fixed it, this time and before) is that compile.js is being ignored completely, unless it also happens to be in a library.

But "." path - the current project is a libray.

> In the current master, this will lead to unexpected side effects - for example, if you run qx compile in your app, then the API viewer's compile.js will be run with the application's compile.json -- but if you run the same command in the API viewer compile.js will be run with the API viewer's .json.  This means that the API viewer's compile.js will be run with completely different (and undefined) source data.

That should be. e.g. apiviewer needs information from the current compile.json: ```    let excludeFromAPIViewer = env.excludeFromAPIViewer;
    let includeToAPIViewer = env.includeToAPIViewer;```


> The other reason that the current master is broken is because the compile.js needs to be able to modify the structure before _mergeArguments is called, because _mergeArguments normalises the configuration as well as integrating command line arguments - for example, it makes sure that there is a target, etc.  However, libraries are not known until after _mergeArguments has normalised the configuration.
> AFAICT the problem which API viewer's compile.js is solving is that it needs to augment the entry for it's application, by updating the application's environment entry; for the API viewer this needs to be done each time the target fires "checkEnvironment", and it just so happens that API viewer ignores the inputData so API viewer is fine with the way that master works.
> IMHO in order to allow libraries to contribute to applications (eg API Viewer) and allow processing of compilation, the API needs to be improved so that there are phases:
> (1) compile.json has been processed (if it exists) and normalisation of the structure has not yet begun;
> (2) normalisation complete, each library's compile.js is asked to contribute
> This will mean that if a library has it's own compile.js used for building it's own demo/test apps, then running qx compile in the library directory will cause both phases to be fired on that library's compile.js
> 

Let's do it in 2 phases:
```
        if (await fs.existsAsync("compile.js")) { 
         config = await this.__loadJs("compile.js", config); 
        }

...
    this._mergeArguments(parsedArgs, config, lockfile_content);
...
	
      if (config.libraries) {
        for (const aPath of config.libraries) {
		  // do not process current compile.js again:
		  if (aPath !== '.') {
             await this.__loadJs(path.join(aPath, "compile.js"), config);
		  }
        }
      }
	
```

hkollmann:
> Something strage here - the PR could be merged even though we don't have two approvals...

We do not have any restrictions yet....

johnspackman:
> But "." path - the current project is a libray.

Only if the current directory happens to be a library (it doesn't have to be), and even if it does it's for a different purpose.

> That should be. e.g. apiviewer needs information from the current compile.json: let excludeFromAPIViewer = env.excludeFromAPIViewer; let includeToAPIViewer = env.includeToAPIViewer;

That's the API viewer adding to the application - but the `compile.json` that is in `compiler.inputData` is different depending on where it is being called from.  So in the current design, the API viewer's `compile.js` cannot rely only anything being in `inputData`, even though that was a kind of a feature.

> Let's do it in 2 phases:

I've been thinking over lunch - the principal has been that `compile.json` is parsed and fed into `compile.js`, and so it would follow that the library's compile.json is parsed and fed into it's compile.js, in order to merge into the application's configuration .... but that's getting quite complicated and doesn't feel right.

So - how about this: there are two distinct types of configuration, ApplicationConfiguration and LibraryConfiguration; these can be classes, one or both of which can be exported from the `compile.js` module.  This could be quite neat because starts to build an actual OO-style API around the pluggable nature of libraries; for example this would lay the groundwork for `qx package install ...` to allow the library to customise the installation (such as when adding an application entry into the `compile.json`)

IMHO it would imply that the compiler ignores `compile.json` if there is a `compile.js` because the ApplicationConfiguration/LibraryConfiguration implementations in `compile.js` would determines what, and when, to load external data.

johnspackman:
> So is the problem that the apiviewer depends on code in `compile.js` which isn't run when [this line](https://github.com/qooxdoo/qooxdoo-compiler/commit/cec91963e924d51a31f03c546844d779664cc780#diff-25f3dd808436d62e36636886c807ba37L120) is removed?

Yes - and it's [this](https://github.com/qooxdoo/qooxdoo-compiler/commit/cec91963e924d51a31f03c546844d779664cc780#diff-25f3dd808436d62e36636886c807ba37R68) line that breaks "normal" configation

> This means we need so kind of "hook" for packages that act as "plugins", don't we?
👍

cboulanger:
> So - how about this: there are two distinct types of configuration, ApplicationConfiguration and LibraryConfiguration; these can be classes, one or both of which can be exported from the `compile.js` module. This could be quite neat because starts to build an actual OO-style API around the pluggable nature of libraries; for example this would lay the groundwork for `qx package install ...` to allow the library to customise the installation (such as when adding an application entry into the `compile.json`)
> 
> IMHO it would imply that the compiler ignores `compile.json` if there is a `compile.js` because the ApplicationConfiguration/LibraryConfiguration implementations in `compile.js` would determines what, and when, to load external data.

Sounds reasonable. To make this BC, the compiler would distinguish between what is imported from compile.js? I.E. function -> current behavior, object having "application" , "library" as members -> new behavior?

johnspackman:
> To make this BC, the compiler would distinguish between what is imported from compile.js? I.E. function -> current behavior, object having "application" , "library" as members -> new behavior?

not quite - because the current version is broken I'd propose that we do not support processing libraries in this way at all.

OTOH a couple of weeks ago, the only projects on GitHub that were using `compile.js` was our code, and we are pre-beta still I think?  We could consider just reporting an error if the new form is not implemented

hkollmann:
We will break all projects which use apiviewer because apiviewers compile.hjs depends on inputdata from the current (!) project.

hkollmann:
Including compiler which is using the apiviewer in this way.

cboulanger:
@hkollmann but if we update compiler and apiviewer together, then they don't break - since we now can specify the dependency. We just need to make sure we develop the solution together.

hkollmann:
you are right - but we have to find another solution than just do not support processing libraries as John suggested.

cboulanger:
@hkollmann  I'd sugget you create a final v0.1 Version of the the API Viewer and tie it to a working compiler version, hen bump the version up to 1.0.0-beta and adapt to whatever we implement now.

johnspackman:
sure, i'll do both compiler and API viewer

cboulanger:
Merging, thanks @johnspackman !

johnspackman:
eek, I thought that the merging would go on hold!  I'm working on it already and aiming to get the PR done this afternoon

cboulanger:
Sorry, I thought you wanted to create a new PR. I have created a "do not merge" label for next time.

cboulanger:
I like the idea of APIfying the compile step. I think it would make it much easier to work with if the  object imported by `compile.js` provides a bunch of specialized methods (such as `getPkgCompileData()` and `getAppCompileData()`) rather than having to work with the compiler object, which might overwhelm people. It could work like a sandbox object which filters access to the compilation process.

cboulanger:
This API could then be extended to provide access to a (virtualiuzed) filesystem, which is important if we want to port the compiler to run in the client.

PR #448, opened 2019-06-14
adds implementation of API for compile.js

johnspackman:
This is a first version, it needs some more testing which I'll do shortly but it's a good showcase.  See also the new commits in API viewer


cboulanger:
Instead of reading `compile.json` from disk, would you consider using the config API? see usage [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/commands/package/Install.js#L520). This makes sure that the file has the correct schema (or will be migrated if possible).

source/class/qx/tool/cli/config/CompilerConfig.js
```diff
@@ -0,0 +1,111 @@
...
+  extend: qx.tool.cli.config.AbstractConfig,
+  
+  construct: function(command) {
+    this.base(arguments);
+    this.__command = command;
+    this.__libraryConfigs = {};
+  },
+  
+  properties: {
+    configFilename: {
```
cboulanger: Is this used anywhere? I think here, too, the config API should be used for consistency.

source/class/qx/tool/cli/config/CompilerConfig.js
```diff
@@ -0,0 +1,111 @@
...
+  
+  members: {
+    __configPromise: null,
+    __libraryConfigs: null,
+    
+    /**
+     * Loads the configuration data
+     */
+    async load() {
+      let compileJsonPath = path.join(this.getRootDir(), this.getConfigFilename());
```
cboulanger: same here: we have a lot of code duplication. See https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/config/Abstract.js

source/class/qx/tool/cli/config/CompilerConfig.js
```diff
@@ -0,0 +1,111 @@
...
+      let config = {};
+      if (await fs.existsAsync(compileJsonPath)) {
+        config = await qx.tool.utils.Json.loadJsonAsync(compileJsonPath);
+      }
+      return config;
+    },
+    
+    /**
+     * Returns the configuration data
```
cboulanger: This returns a promise, doesn't it? This should be documented...

cboulanger:
@johnspackman  I wonder: with the `LibraryConfig` and `CompilerConfig` classes, what are you actually modeling? I am a bit worried we duplicate `qx.tool.config.Abstract` and its subclasses here. The methods of `CompilerConfig` could actually be integrated into `qx.tool.config.Compile`. `LibraryConfig` is different, but the `load()` method does not actually refer to loading library data (which would be the equivalent of loading compiler config data), but is just a hook for running library-specific code. I understand what you're trying to do, but it doesn't seem completely right and I wonder if there was a more intuitive and elegant solution for the real problem we are trying to solve.

cboulanger:
I am all for returning a class/classes from compile.js which have a proper API. But I wonder if rather than having two separate classes, if a "sandbox object" would be easier to reason about, which would have a unified, easy API and which would contain necessary data, forward lifecycle events from the compiler object so that libraries can run code and manipulate config data.

source/class/qx/tool/cli/commands/MConfig.js
```diff
@@ -116,6 +137,28 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
       }
       this._mergeArguments(parsedArgs, config, lockfile_content);
 
+      if (config.libraries) {
+        for (const aPath of config.libraries) {
+          let compileJsFilename = path.join(aPath, "compile.js");
```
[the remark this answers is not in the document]
johnspackman: the `compileJsFilename` here is different to the one above - this is the name inside the library and has to be hardcoded, the one above is derived from the command line and can be overridden.  We have to have a fixed name across all libraries that can;'t be overridden on the command line.  I take your point about a constant, I've added that for the next commit

source/class/qx/tool/cli/config/CompilerConfig.js
johnspackman: It's used in the default implementation, and it says what the compiler thinks the .json file is should be called; this can be overridden by the command line args which is why it is passed in here.  The config API has been used as part of determining what this should be (in MConfig)

source/class/qx/tool/cli/config/CompilerConfig.js
johnspackman: The only duplication is that the default implementation loads a file via a call to a utility function - `qx.tool.config.Abstract` has a *lot* more code in it which does not apply here

johnspackman:
What's being modelled is an API to allow the application and libraries to interact with the compiler, and currently the only mechanism is to allow them to find alternative ways to generate a configuration.  For many applications, that just means loading a file but could be returning an object etc; for libraries like API viewer, that means adding environment settings but could also involve loading files.

WRT the config API, I don't think there is a lot of overlap - although I think that it would be helpful if the package name was "API" rather than another "Config".  

`CompilerConfig` class does not use the config API is because it is just a means to generate a config, which by default is just by loading a `compile.json` file but that is not necessarily true - for example, it may generate one without loading anything (which could be a common use case, you wouldn't necessarily want to keep your static JSON data separate from your code that manipulates it).

File which are loaded do not have to confirm to the `compile.json` schema (or any other schema), but what it outputs *must* comply.  There's no reason why, for example, the configuration can't be invalid (from a compile.json schema perspective) until after all of the libraries have had their opportunity to add to the configuration.

The schema validation still happens elsewhere in the compiler, I havn't changed that - but it happens on the final output and not intermediate fragments

johnspackman:
> I am all for returning a class/classes from compile.js which have a proper API. But I wonder if rather than having two separate classes, if a "sandbox object" would be easier to reason about, which would have a unified, easy API and which would contain necessary data, forward lifecycle events from the compiler object so that libraries can run code and manipulate config data.

I don't really follow, the reason that the CompilerConfig and LibraryConfig are separate is because there can be multiple LibraryConfig's and one CompilerConfig; this is modelling the state which contributes to the final configuration.  We could merge it into one class, but then the individual library's `compile.js` would have to export some kind of state or functions to perform various tasks, whereas this way they just extend an OO class as usual.

For example, `qx package install ...` manipulates the application's compile.json in order to add applications to the `applications[]` array, but it might be useful to allow the library to choose how it's added so that special settings for theme or environment can be injected.  Similarly, installing a library via `qx` might makje choices (including asking the user questions) about merging theme's etc.  To implement any of that kind of thing, `LibraryConfig` would have methods added which provide default implementation, and the library's `compile.js` can provide a class derived from `LibraryConfig` which overrides the method.

I'm changing the names in the next commit to API, so maybe that makes more sense?

cboulanger:
@johnspackman Ok, we're probably solving different problems. I am deferring on your judgement here. However, we have to document the solution well in the documentation and the API Viewer, because I didn't understand it the first time.

johnspackman:
I think you're right that it is a different problem - effectively, this sets the groundwork for a generic API to the CLI, from the point of view of compiling a project (ie where you run `qx compile`) and the libraries (which can, but does not always, include ".").  And the only API method that is implemented so far is a glorified replacement for `qx.tool.util.Json.loadJsonAsync("compile.js")`, and schema is added later.

Documentation is written and in https://github.com/qooxdoo/qooxdoo/pull/9701, hope it makes more sense now.

Also, when writing the docs it occurred to me that libraries don't need to have code for the project being compiled to be useful - [see here for the SASS compiler example](https://github.com/qooxdoo/qooxdoo/blob/d6bc13d3cdd05a4b6e3cdb9eca047d5d31d42617/docs/configuration/api.md#reusable-functionality-through-code-less-libraries)

PR #449, opened 2019-06-16
Link the "documentation" link on the `qx serve` landing page to the documentation folder in the framework

cboulanger:


cboulanger:
@hkollmann You mean [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/resource/qx/tool/cli/serve/src/index.html)?

hkollmann:
Found in partials/footer.html.

cboulanger:
Fixed the links in the `src` folder also. 

I also added a  `--show-startpage` / `-S` parameter to `qx serve` so that the documentation link can be accessed even if there is only one application.

cboulanger:
@johnspackman, can you have a look?

johnspackman:

PR #451, opened 2019-06-16
fix building of the website for qx serve

hkollmann:


johnspackman:
merging thank you @hkollmann !

PR #453, opened 2019-06-17
exposes the `.getMaker()` method publicly and adds missing tests

johnspackman:
Apologies for the late entry on this one, I intended for this to be part of the last PR.

The examples call the `._getMaker()` method of the command in order to integrate fully with the compiler API, but this is not ideal given that it is a protected method.  The maker is an essential entry point, so this change exposes it publicly.

Once this is merged, I'll update API viewer and Demo browser to match

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "puppeteer-to-istanbul": "^1.2.2",
     "replace-in-file": "^4.0.0",
     "rimraf": "^2.6.2",
+    "runscript": "^1.3.1",
     "semver": "^5.7.0",
     "set-value": "^3.0.0",
     "tmp": "0.0.33",
--- a/package.json
+++ b/package.json
@@ -75,7 +75,6 @@
     "puppeteer-to-istanbul": "^1.2.2",
     "replace-in-file": "^4.0.0",
     "rimraf": "^2.6.2",
-    "runscript": "^1.3.1",
     "semver": "^5.7.0",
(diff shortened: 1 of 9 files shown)


test/testapp/compile.js
```diff
@@ -19,11 +21,64 @@ qx.Class.define("qxl.compilertests.testapp.LibraryApi", {
...
+    __startedSassWatch: false,
+    
     async load() {
       let command = this.getCompilerApi().getCommand();
+      command.addListener("making", () => this._onMaking());
       command.addListener("checkEnvironment", e => this._appCompiling(e.getData().application, e.getData().environment));
     },
     
+    _onMaking() {
+      debugger;
```
cboulanger: Is this here on purpose? I haven't used this in a while, but somehow remember it would stop execution...

johnspackman: it only stops (pauses) under the debugger, but it shouldnt be there - removing it now

hkollmann:
@johnspackman: Could you resolve the conflict and have a look at my Question please?

package.json
```diff
@@ -75,6 +75,7 @@
     "puppeteer-to-istanbul": "^1.2.2",
     "replace-in-file": "^4.0.0",
     "rimraf": "^2.6.2",
+    "runscript": "^1.3.1",
```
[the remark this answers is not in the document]
johnspackman: done

Issue #450, opened 2019-06-16
qxcompiler can not compile itselfs any longer


hkollmann:
How to reproduce:

git clone https://github.com/qooxdoo/qooxdoo-compiler.git
cd qooxdoo-compiler
npm install
./cqx compile

get's 
Error: TypeError: fixLibraryPath(...).then is not a function
    at qx.tool.utils.files.Utils.correctCase.then.then.then.data (/mnt/c/_vza/@repos/github.linux/qooxdoo-compiler/source/class/qx/tool/compiler/app/Library.js:174:14)

johnspackman:
There is [a bug here in app/Library.js](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/app/Library.js#L160) where the code should return `qx.Promise.resolve(dir)` instead of just `dir` - I'll submit a PR for that in a moment because another issue that missing source or resource dirs are not handled very elegantly.

However, the underlying cause is a bigger problem - the missing files are `node_modules/@qooxdoo/framework/{source,resource}`, but this is because the directory structure is now `node_modules/@qooxdoo/framework/framework/{source,resource}` - is the build process for `@qooxdoo/framework` broken?

johnspackman:
Something odd that's happening on my Mac is that I although the directory structure is the same as on my Windows VM, `./qx compile` works - I'm presuming for the moment that is something to do with my dev setup

hkollmann:
I will check this.

hkollmann:
qooxdoo sources should be in node_modules/@qooxdoo/framework/framework/{source,resource}.
node_modules/@qooxdoo/framework is the root of the package and contains tool, lib, doc and framework folder.

hkollmann:
We could/should change this during building. We can copy framework/source to source and include source to the npm file. This would make the package structure much more clearly.
What do you think?

johnspackman:
yes I agree - there is currently a `node_modules/@qooxdoo/framework/Manifest.json` *and* `node_modules/@qooxdoo/framework/framework/Manifest.json` which is confusing, but IMHO either the first one needs a correct path in it's `provides.class` or the whole tree should be moved up (whcih would be preferable) and then `qx`'s auto-detect for the framework would work just like it does for any other directory which contains a library

hkollmann:
I will prepare a PR for deploy_npm in the evening

hkollmann:
done with #9707

PR #454, opened 2019-06-17
Fixes crash and improve console output

johnspackman:
This fixes part of the problem with https://github.com/qooxdoo/qooxdoo-compiler/issues/450, where a missing library `class` directory causes an odd error message; it also tries to improve the console output.

Previously, there were two log appenders, one was `NodeConsole` which output everything and the other is `qx.tool.cli.LogAppender` which outputs only if a defined level is used, and also outputs date/time information with each entry.  The result is a lot of duplication of output and extra noise in the form of date/time data (down to the millisecond).  This PR reduces it to just `LogAppender` and removes date time output

### DIFF
--- a/source/class/qx/tool/cli/Cli.js
+++ b/source/class/qx/tool/cli/Cli.js
@@ -29,11 +29,13 @@ qx.Class.define("qx.tool.cli.Cli", {
 
   members: {
     run: function() {
+      /*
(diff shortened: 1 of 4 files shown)


cboulanger:
This fails because of an overly aggressive lint rule - we should demote the “no-unused-vars” rule to “warning” @hkollmann

Issue #452, opened 2019-06-17
Add option to qx serve to build local helper apps or link qooxdoo website


cboulanger:
`qx serve` should have an option that allows to either make all links on the landing page to the qooxdoo website (dependent on the version that it depends on) OR to build all necessary apps (API Viewer, Playground, WidgetBrowser, etc.) locally and set the links to them. This will turn the landing page into a really useful resource in addition to letting users access their applications.

cboulanger:
I wonder: shouldn't we just rewrite the [qooxdoo website build script](https://github.com/qooxdoo/qooxdoo/blob/master/.travis/build-site) using compiler/CLI and use that script both for the website AND for the local website? The only thing it would differ in is that the API viewer would contain application classes and the middle part of the site should contain some nice promotional texts and shiny images instead of the list of applications. The infrastructure is already in place since the HTML is generated from templates.

oetiker:
hihi that would be neat ... then we just have to add 'version handling' to the online version

cboulanger:
@oetiker Interested to make this happen? I'd love to have a new website as soon as possible, and it seems to be feasible with what we already have, since the Metalsmith templates seem to allow for a lot of flexibility. We could control which kind of website (local "qx serve" or public website) via environment variables.

cboulanger:
Implemented, closing

PR #456, opened 2019-06-18
Fix and update script which generates local webserver landing page

cboulanger:
This fixes a broken path and "modernizes" the code to use async syntax in preparation of #452


johnspackman:
Merging thank you @cboulanger

PR #458, opened 2019-06-19
Do NOT just install the current version of @qooxdoo/framework.

oetiker:
the current version of the framework is nice for the current version of the compiler, but when running an old compiler this messes up things badly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "main": "index.js",
   "scripts": {
     "test": "cross-os test.os",
-    "postinstall": "npm update --no-save @qooxdoo/framework",
     "compileWebSite": "cd source/resource/qx/tool/cli/serve && node compile.js",
     "lint": "eslint lib"
   },


hkollmann:
Seems we should change the deploy_npm script to update the package.json and package-lock in the moment the new package is build. This would help both. You have a stable combination of compiler/framework and we can not forget to update the version in package and package-lock.

oetiker:
maybe it would suffice to just add an `npm update @qooxdoo/framework` to the packaging and shrinkwrapping process. BEFORE runnig the tests.

hkollmann:
That‘s one Part. The Second is to change postinstall to beforepublish so that the Package is updated when you run npm install


Von meinem iPhone gesendet

hkollmann:
@oetiker : Could you have a look at this https://github.com/qooxdoo/qooxdoo-compiler/pull/459? Should match better then your PR.

PR #459, opened 2019-06-20
Fix install mess

hkollmann:
- move postinstall to prepare. So update will run only during npm install
- add prepack script to bind framework version to compiler
- fix some issues in deploy_npm
- add .npmignore. The .travis folder should not be part of the npm package
- run npm shrinkwrap to get actual shrinkwrap file
- remove git dependency


hkollmann:
As i promised, here is a replace for https://github.com/qooxdoo/qooxdoo-compiler/pull/458.
This should adress both problems. Have a stable package and update framework during install from GitHub and pack for npm.

oetiker:
@hkollmann my npm foo is weak, so I can not tell from the code ... BUT will the compiler be tested WITH the latest qooxdoo version already in place?

hkollmann:
Good hint. Just tested npm link. prepare runs during link process - and test script runs npm link. So Framework will be updated before test

hkollmann:
Here is the link for npm scripts: https://docs.npmjs.com/misc/scripts

johnspackman:
is this ready for merge?  @oetiker please can you confirm that your question is answered?

PR #460, opened 2019-06-20
fix paths for doc and manifest.

hkollmann:


cboulanger:
still errors: [Error: Empty Manifest.json in library at /home/travis/build/qooxdoo/qooxdoo-compiler/node_modules/@qooxdoo/framework/framework]

PR #462, opened 2019-06-22
APIfy website generation

cboulanger:
In preparation of #452 , this moves the build scripts for the local website into the `qx.tool.utils.Website class` to make them callable from the CLI code.


cboulanger:
This removes the `qx/source/resource/website/build` folder and builds the local website on demand the first time it is used. It can be rebuilt using the `qx serve -R` option. Once @hkollmann adds statically the built qxl.* apps to the NPM package, we can switch to linking to them from the local web page.

cboulanger:
@johnspackman can you elaborate?

hkollmann:
Great! We should think About if if use the builder construction for all commands. But not in this PR.

cboulanger:
@hkollmann You're right, for example `-v`/`--verbose` should be inherited rather than repeated in each builder... can you put this into a separate issue?

johnspackman:
> @johnspackman can you elaborate?

if the compiler version number changes, the compiler will delete the output directory before starting the compile (that link to the Maker class was to illustrate where).  IMHO a similar detection for the compilation, when the version number changes it gets recompiled.

OTOH, unless there is a need to compile it with a user's project code, I don 't understand why this does not remain part of the deployment of new qx versions.

cboulanger:
@johnspackman The reason I removed the build was that 1) it seemed strange to me to change the same things at two places: in the source and the build version, and 2) that I thought that it would be good to be able to dynamically change the local website based on specific configurations (for example, to link to locally built playground versus the one on the qooxdoo website). Since then, @hkollmann has rightly suggested that it makes more sense to include those apps pre-built in the NPM package - and I think that applies to the local website as well. Would you support this solution - to make website building part of the NPM deployment?

johnspackman:
@cboulanger yes, I think inclusion in the npm deployment is ideal

johnspackman:

PR #465, opened 2019-06-23
For applications based on packages you must provide the full path to …

hkollmann:
…the package for the boot path


johnspackman:
merging, thank you @hkollmann

PR #466, opened 2019-06-23
Use local assets for website instead from CDN

cboulanger:
The `qx serve` website should use locally cached assets instead of fetching them from CDNs so that it can work offline. This PR adds this. Depends on #462.


source/class/qx/tool/cli/commands/Serve.js
```diff
@@ -34,105 +34,38 @@ qx.Class.define("qx.tool.cli.commands.Serve", {
 
   statics: {
 
+    builder: {
```
johnspackman: I think there's a style guide somewhere that says that static values are supposed to have upper case names?  Also it would be good to see this named to show that it's for yargs, eg `YARGS_BUILDER` rather than just `builder`

cboulanger: I'll fix this in #462 , on which this PR is based.

cboulanger: Fixed in https://github.com/qooxdoo/qooxdoo-compiler/pull/462/commits/5b5a53d4efa05dbefb35f07532b28a9aec938614 , please approve & merge that PR first.

source/class/qx/tool/utils/Website.js
```diff
@@ -0,0 +1,240 @@
...
+      metadata.site.pages = pages;
+    },
+
+    /**
+     * Metalsmith plugin that loads partials and adding them to the metadata.partials map.  Each file
+     * is added with its filename, and if it is a .html filename is also added without the .html
+     * extension.
+     *
+     */
+    loadPartials: async function (files, metalsmith) {
```
oetiker:
```suggestion
    async loadPartials (files, metalsmith) {
```

cboulanger:
I've implemented all requested changes, and tests pass. PR can be approved & merged now.

source/resource/qx/tool/bin/download-assets
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
```
hkollmann: Could we have a node script here? Node is running on Windows too, bash not directly.

cboulanger: Could you add one later? The assets are committed statically anyways, so this script needs to be run very rarely  to update the scripts.

cboulanger: (You can download and replace individual assets manually if the need arises, this script was just for convenience)

cboulanger:
Can this be approved now?
We need a similar solution for the documentation (see https://github.com/qooxdoo/qooxdoo/issues/9705) but it cannot be solved together because of the separate repos.

johnspackman:

PR #467, opened 2019-06-25
2 changes:

hkollmann:
- only list apps which are build for the browwser (qooxdoo server can not run in the browser!)
   - serve  /apps


johnspackman:
merging, thank you @hkollmann

PR #468, opened 2019-06-25
`qx package upgrade -p` will now upgrade to the latest prerelease

cboulanger:
Before, it was not possible to upgrade to a prerelease.


johnspackman:

PR #470, opened 2019-06-25
add apps to server

hkollmann:
fix special case for serving qooxdoo farmework


johnspackman:
merging, thank you @hkollmann

PR #471, opened 2019-06-25
move to newest apiviewer to get rid of warnings

hkollmann:


cboulanger:
Merging, assuming @johnspackman's approval since this is just an update of dependencies, thank you @hkollmann !

PR #472, opened 2019-06-26
adds console colorisation

johnspackman:
allows the default color for console output to be set with (eg) `qx config set qx.default.color 'white bgRed bold'`

default value for --no-feedback is set with `qx config set qx.default.feedback false` (if not specified, default remains "on" as before)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1593,6 +1593,7 @@
         "anymatch": "^2.0.0",
         "async-each": "^1.0.1",
         "braces": "^2.3.2",
+        "fsevents": "^1.2.7",
         "glob-parent": "^3.1.0",
         "inherits": "^2.0.3",
         "is-binary-path": "^1.0.0",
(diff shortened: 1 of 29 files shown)


cboulanger:
es-lint is not nice to you, it seems 😄  But probably `qx lint --fix` should help.

johnspackman:
oops, one day i'll remember to run lint before i commit! :)

derrell:
Just saw this. Can you confirm that color can be turned off entirely in a config file? Color doesn't work in some environments, yielding instead a bunch of control characters visible on the screen...

johnspackman:
yes - the colours only apply if turned on via the config setting, otherwise they are as before.  

However, the feedback option (which is on unless `--no-feedback` is given or `qx.default.feedback` is set to false) will use ANSI sequences including colour - this is also as before, but I imagine that these would cause the same issues as colour?

derrell:
Yes, that would cause the same problem. IMO, either it should look at your TERM environment variable and determine that you have color available or are an ANSI-compatible "terminal" type, or not use coloring unless specifically asked to.

johnspackman:
It should do - I've used the [console-control-strings](https://www.npmjs.com/package/console-control-strings) npm module because that's the same module which the gas gauge module uses for all the pretty termcap-esque progress bar

PR #473, opened 2019-06-26
Update status badges in readme

cboulanger:


johnspackman:
merging despite failures because it's doc only and the failures are unrelated; thank you @cboulanger

cboulanger:
Doesn't seem to work, though - still displays alpha version ...

PR #477, opened 2019-06-29
fix the links in website

hkollmann:


cboulanger:
Looking forward to that one, thanks @hkollmann !

PR #478, opened 2019-06-29
new version of qx-rules to speedup no-illegal-private-usage

hkollmann:


cboulanger:

PR #481, opened 2019-06-30
fixes for @qooxdoo/qx/no-illegal-private-usage rule

hkollmann:


hkollmann:
You are right. It's not a qooxdoo class so just ignore "@qooxdoo/qx/no-illegal-private-usage"

johnspackman:
thank you @hkollmann

PR #483, opened 2019-06-30
Fix a typo and add demobrowser to devtools

cboulanger:


johnspackman:
thank you @cboulanger

PR #484, opened 2019-07-04
to compile test code there are 2 fixes necessary:

hkollmann:
- classname is allowed to be null, see API doc:
                       qx.Class.define: name (optional; default: null) Name of the class.
                       If null, the class will not be added to any namespace which could be handy for testing.
   - extend is to be allowed null: test: extend from null should extend Object


johnspackman:
merging, thank you @hkollmann

PR #485, opened 2019-07-05
website generation is local - so move metalsmith packages to install …

hkollmann:
…section.

Without npm install @qooxdoo/Compiler is not runable


johnspackman:
merging, thank you @hkollmann

PR #491, opened 2019-07-15
Adds info metadata to qx list internal indexes

cboulanger:
also fixes gitignore files


cboulanger:
I'll push a change that will store the whole manifest instead of just the `info` part. Doesn't make sense to save just a part of the metadata, since it's all local in-memory stuff.

johnspackman:

Issue #489, opened 2019-07-11
Add support for targets for different environments


johnspackman:
When compiling multiple applications only one "source" and one "build" target are possible (or required); each target shares the same `transpiled` directory between all applications, which means that environment settings and babel settings are common to all applications.

Babel settings are global - it is not possible to set babel settings specific for a single application because babel settings affect the code generated in `transpiled/`.

This means that if you have a node server application and a web browser application, you have to use the lowest common denominator - ie even if you can guarantee that your node version is the latest and greatest, all code is transpiled as though it has to run on the oldest web browser.

For environment settings, settings such as `qx.debug` are used to eliminate code provided that the environment value is identical for every application; it is not possible to target specific environment settings for individual applications *and* simultaneously have static code elimination.

I propose we allow multiple targets of a given type, and add to the `target` definition in `compile.json` to support two new properties:

* `application-types` - an array of application types (eg `"browser"` or `"node"`) which this target supports.  Optional, and if not provided would mean that all types are supported.
* `application-names` - an array of application names (eg from `applications[0].name`) which this target supports.  Optional, and if not provided would mean that all applications are supported.

The compiler would be able to identify one or more targets for source or build, and then generate all the applications which match the `application-types` and `application-names` filters.

hkollmann:
In the current version you can do this with multiple compile.json files and different target dirs. I use this in qooxdoo to distinguish between server compile and app compile.

johnspackman:
You can have multiple processes and define your applications separately, but the proposal is to eliminate the need to do that; in my case, those multiple `compile.json` would be 99% identical and have to be maintained (unless a custom compile.js is built which includes data).  

It's not just DRY that motivates this, logistically it would be more cumbersome because my builds have a rhino app, a node app, and at least one browser app - so that would mean three terminal windows running `qx compile --watch`.   That's per website, and in one instance there are three websites which are part of the same deliverable system, so that would mean up to 9 terminal windows each running `qx compile --watch`.

cboulanger:
@johnspackman If you can invest the time to implement the proposal, that would be great.

hkollmann:
Done with #493

PR #493, opened 2019-07-15
adds support for multiple targets

johnspackman:
https://github.com/qooxdoo/qooxdoo-compiler/issues/489


source/resource/qx/tool/website/partials/head.html
```diff
@@ -15,6 +15,7 @@
     <title>{{= it.title || it.site.title }}</title>
     <meta name="description" content="{{= it.excerpt || it.site.excerpt || "" }}">
     <link rel="stylesheet" type="text/css" href="/qooxdoo.css">
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.5.5/bluebird.js"></script>
```
cboulanger: It would be better to store the file locally and load it from there so that it works in offline situations.

johnspackman: done

hkollmann:
Ready for merge?

johnspackman:
@cboulanger already written, pushing now :)

@hkollmann yes please 👍

hkollmann:
Thanks @johnspackman ! Great work!

Issue #496, opened 2019-07-16
top-level index.html sets qx.$$appRoot incorrectly


cboulanger:
If no custom `boot/index.html` is specified, the `index.html` generated in the top-level folder contains a script tag which sets `qx.$$appRoot` to the name of the subfolder containing the main application. This leads to an incomplete URL (no protocol, no host, etc.) when using `qx.$appRoot` directly or via the ResourceManager.getUri() API. 

After removing the script tag manually, qx.$$appRoot is calculated correctly. However, on the next compile, it will be regenerated. 

Probable fix: remove the script tag from the template for the top-level index.html.

hkollmann:
qx.$$appRoot holds the relative path from the root directory into the app dir if we have root application.
Otherwise - if we have no root app - it's empty.

So this is a feature - not a bug.

cboulanger:
The problem is the of non-deterministic (from the perspective of the application) generation of ResourceManager.toUri() - if `qx.$$appRoot` is manually set, it results in a relative path, if not, the result is an absolute URL. Be that as it may, If removing the script tag from the top-level index.html fixes the issue, we should remove it, no?

johnspackman:
I think that the issue is that we can't just remove it, it exists because in environments where the root of the web server is the target output directory, http://www.myserver.com requires that script tag/$$appRoot.  

(It may be that in this example, http://www.myserver.com/index.html is actually OK because you can't reduce the path any further)

However, AIUI if the root of the web server is elsewhere (eg the application directory) then http://www.myserver.com/compiled/source also needs that script tag but http://www.myserver.com/compiled/source/index.html does not, even though they actually load the same file.

Another thing to note is that http://www.myserver.com/compiled/source and http://www.myserver.com/compiled/source/ are different, but again load the same index.html file.

I think that the detection in the loader.browser.tmpl.js needs to be improved

PR #501, opened 2019-07-20
fix for https://github.com/qooxdoo/qooxdoo-compiler/issues/496: top-l…

hkollmann:
…evel index.html sets qx.$$appRoot incorrectly.

There was an improved version in the template, i just activated it.


cboulanger:
Yes, let's omit the generation of qx.$$appRoot in the standard template and improve the detection code instead.

hkollmann:
closed because of #507

PR #506, opened 2019-07-25
This adds optional "description" and "showInPackageBrowser" properties to compile.json

cboulanger:


cboulanger:
This is needed for the package browser.

hkollmann:
not really. If you do not publish the app as package it will not be listed in the packagebrowser.

hkollmann:
if we really need a attribute for the this purpose should we add it to manifest and give him a more general name, something isPublic or so

cboulanger:
@hkollmann Please note: this is not about the library (=Manifest.json) but about one of the demo applications, of which the metadata exists only in `compile.json`. These applications are only built locally or in the package browser. This property prevents the demo app to be published in the package browser, which make sense if the demo works does not work standalone (, i.e.only in a local environment witha server etc.) or is an internal application which should not be made public.

This is different and separate from apps that are published in `Manifest.json`.

cboulanger:
With "should not be made public" I don't mean that they are protected somehow (they are not, since anyone can clone the repo and build them), but that they might be , for example, test apps that don't add anything to the public profile of the package.

PR #508, opened 2019-07-25
Rename "showInPackageBrowser" to "publish"

cboulanger:
ok let's try this


johnspackman:
thank you @cboulanger !

PR #509, opened 2019-07-25
the authors key is an array ...

oetiker:


cboulanger:
eslint is complaining about one-liners in the code. 
Off-topic here but does this rule make sense?

hkollmann:
@cboulanger : With function closed {} yes - it's more readable.

cboulanger:
Doesn’t qx lint --fix help?

hkollmann:
Before run qx lint --fix you must run:
qx pkg update
qx pkg install

cboulanger:
Merged, thank you @oetiker !

PR #512, opened 2019-07-26
Resource manager fixes

johnspackman:
This changes the SASS compiler so that it is either the old-style or the new-style, globally.  By default it is old-style but if not explicitly chosen then the first compilation will issue a deprecated warning.  The idea is that in v7 the old style will disappear.

compile.json now supports `sass.compiler` which can be "legacy" (to suppress that deprecated warning) or "latest" to switch on the new style.

This also sneaks in a couple of other warning fixes


johnspackman:
The qxl.mobileshowcase has been updated to match, it's reverted the moving of scss files

cboulanger:
Merging, thank you, @johnspackman !

Issue #514, opened 2019-07-29
config.js of package is not executed


hkollmann:
If packages are loaded during compile the config.js in the fresh loaded package is not executed.

how to reproduce:

git clone https://github.com/hkollmann/hkollmann.mobilegrid.git
cd hkollmann.mobilegrid
npm install @qooxdoo/Compiler
npm qx compile -v
run again

In the second log you see that the apiviewer is analysing the classes. In the first run not.

hkollmann:
fixed with #515

PR #515, opened 2019-07-29
load missing libraries earlier

hkollmann:
fix for #514 

This will fix error of running lint without loaded libraries too


hkollmann:
BTW: That's the reason while config.js in packagebrowser is not executed!

cboulanger:

PR #516, opened 2019-07-29
Compiler fixes

johnspackman:


hkollmann:
I would prefer my solution for #514 because your solution did not fix the problem of missing libs during lint.

johnspackman:
yes, I agree - i'll rebase once your PR is fixed

johnspackman:
This PR now incorporates https://github.com/qooxdoo/qooxdoo-compiler/pull/515

hkollmann:
You delete all the testdata stuff - but now the tests are not running  anymore?

PR #518, opened 2019-07-30
Fix startpage if getWriteIndexHtmlToRoot() is set

hkollmann:


source/class/qx/tool/cli/commands/Serve.js
```diff
@@ -112,7 +112,7 @@ qx.Class.define("qx.tool.cli.commands.Serve", {
 
       const app = express();
       const website = new qx.tool.utils.Website();
-      if (apps.length === 1 && apps[0].getWriteIndexHtmlToRoot() && this.argv.showStartpage === false) {
+      if (((apps.length === 1) || apps[0].getWriteIndexHtmlToRoot()) && (this.argv.showStartpage === false)) {
```
johnspackman: this isn't quite right - it may be `apps[1]` that has `.writeIndexHtmlToRoot` set

johnspackman:

Issue #522, opened 2019-07-31
source code formating


johnspackman:
From https://github.com/qooxdoo/qxl.apiviewer/issues/16

> @oetiker I like adding little examples to my qooxdoo class jsdoc:
https://github.com/oposs/headerpuppet/blob/master/source/class/headerpuppet/HeaderPuppet.js#L20
when using `<pre class="code">` indentation does not work

hkollmann:
fixed with #523

PR #523, opened 2019-07-31
fix for #522

hkollmann:


johnspackman:
@hkollmann there's a lint failure

PR #526, opened 2019-08-01
Show how to initialize a directory and install the compiler

derrell:
These initial steps allow one to start with no knowledge, and build a running app.


cboulanger:
I am not against this change, but wasn't the idea to keep this as simple as possible? With `npx` you don't even need to install anything into the project directory.

derrell:
I think I didn't get the point of this PR across. The point is that the testdrive doesn't work, if you don't have the compiler installed, yet there is nothing in the instructions that tell you how to install the compiler. Yes, one could clone the repo, but that's not what we expect people to do. Rather, we expect them to install the compiler via `npm`. It is not at all intuitive how to do that. You have to go learn how to search for something using `npm`, and even once you figure out how to do that, you find about 20 entries when you do `npm search qooxdoo`, so which of these do you install?

So unless you're telling me that `npx qx create myapp --noninteractive` will go out and figure out how to install qx so that it can create myapp, I still think this is critical information to have on the page.

As to `npm init -y`... Maybe `npx` does that for you? Somehow, you need a package.json file in your directory, or npm will install packages in the first higher directory it finds that has a package.json.

oetiker:
maybe pointing to http://www.qooxdoo.org/6.0.0-beta/docs/#/?id=getting-started would also serve

johnspackman:
I agree that this needs to be fixed, but I'm with @cboulanger in that something simpler would be ideal - I know that there are objections to installing globally and that this is "bad practice", but to get potential new users up and running, the simpler the better.

How about this:

```
$ npm install -g @qooxdoo/compiler
$ qx create mydemoapp --noninteractive
$ cd mydemoapp
$ qx serve

```

derrell:
Sigh. I just hate the idea of teaching people to do something that's not the way you think they should really do it, just for expediency. It's really pretty inappropriate for us to tell 'em to install it globally, but then when they are talk with us on gitter, we recommend that they install it locally with each project they work on.

Showing how to install it globally is better than not showing anything about it, since it shows the package that needs to be installed. Maybe just add a comment that for this quick testdrive, we'll install it globally, but for real projects, we recommend that it be installed as part of the project?

johnspackman:
IMHO the "good practice" of not installing globally is debatable and while I can see the benefits of installing a specific version for a project, I'd guess that this is most useful when working in a team so everybody gets exactly the same results, bugs and all.  When a project switches into long-term support that's another time that it *may* be useful to attach a specific version of the compiler, just to make absolutely 100% sure that future minor maintenance work doesn't have unexpected consequences.

But the rest of the time I personally don't agree that it is useful to install per project; if you have multiple projects on the go, you're probably more interested in them all being on the same version for example.

And it kind of goes against our promise of backwards compatibility - if it's such a major concern to us, why are we recommending that the developer lock in an exact version into their code base?

For example, if someone reports a problem the first thing we're going to suggest is try it on the latest version of the framework and the toolchain

What would possibly be more useful is to record the version of the compiler used in the `qx-lock.json` file, and then issue a clear warning if the current compiler version is older to that recorded in the lock file.  All versions are stored in perpetuity in npmjs.com so you can always switch to that older version locally to the project if you need to, at a later date.  As a team you know you're always on the same version and similarly if you're getting an old project out of mothballs you identify the same potential issue.

I'm entirely guessing based on my own meandering experience but I suspect that this will work better for the majority of use cases *and* be a reliable solution.

cboulanger:
I agree with John that there is no single best practice that is right for everybody. I am a "one-person team" that works on apps which I maintain just by myself, and I work with a global compiler that I upgrade as updates come in. But that's mainly because I need the newest features. I completely understand that that's not the right way for the kind of projects @derrell and @oetiker work on. But be that as it may, a "test-drive" in my book is just that - a fast, disposable pile of bits which are solely used to  try out a library. But we should by all means point that out and explain that we recommend a different approach that can be explained somewhere else in the documentation.

derrell:
Ok, I can buy into it. In testdrive, we have them install qx globally, but point them to the getting-started doc which discusses both options. 

What does npx do with all of the package files that need to get installed if you don't do `npm init` to create a package.json file first? I'm just concerned that we clutter up some other node_modules directory with stuff from qooxdoo, because of the normal search-upwards-until-you-find-one standard policy.

cboulanger:
As per https://www.npmjs.com/package/npx , if you provide the `-p / --package` option to `npx`, you can avoid installing anything (globally or locally) completely, as it will "use a freshly-installed, temporary version of the package". So the correct code would probably be 
```
npx -p @qooxdoo/compiler qx create foo
cd foo
...
```
What I don't know (and cannot test ATM) is whether the package has to be specified again the second time `npx` is used (which would defeat the purpose) or not.

cboulanger:
@derrell What's the state of this PR - do you still want to make a change to the current docs in this regard?

derrell:
We don't have consensus on global vs local installation of `@qooxdoo/compiler`. I still think that we should mention both global and local installation here, with reasons you might want to do it one way or the other... or link to someplace that does so. If others still think that's too much for this section, I would close out this PR.

cboulanger:
The current docs are incorrect because `npx` has to be executed in a directory that already contains a `node_modules` folder with the compiler code. So @derrell's patch is correct. The only way using `qx` without installing anything permanent would be to execute it  as `npx -p @qooxdoo/compiler qx ...` which will download the compiler every time you execute a command. Therefore, we need to either accept @derrell 's change or change the docs to say that a global `qx` should be used.

hkollmann:
@johnspackman : Could we merge this?

johnspackman:
I would give the option of installing globally first - IMHO if you're concerned about installing globally, you'll go looking for that option...but if you just want to have a play and see how simple it *can* be, the global option should be first because it is so much simpler.  

It's almost that it is not worth mentioning the install locally version - lots of packages only advertise the global option, but we all know, as experienced developers, that we could make it work without global whatever it is.  IE anyone who cares...already knows how to install locally.

FWIW I took this approach in the overview document I did recently
https://github.com/qooxdoo/qooxdoo/tree/master/docs/development/compiler

derrell:
I was sufficiently happy with what @johnspackman did in https://github.com/qooxdoo/qooxdoo/tree/master/docs/development/compiler. I would have preferred the other order -- do a local install, initially, and then explain how to do a global install if you really wanted to -- but both were provided, so I was happy with it.

Maybe here, we do the global install (but we must actually have the global install statement there, not just assume it's already installed), and then point to that document for additional information.

Regardless, it's broken without a change one way or the other. Obviously I prefer my PR's approach... :-)

README.md
```diff
@@ -48,8 +48,18 @@ nvm use 8 # or 10
...
+cd testdrive
+npm init -y
+npm install @qooxdoo/compiler
+```
+
+Now, create your testdrive application using `npx` which doesn't install anything
+globally.
 
 ```bash
 npx qx create myapp --noninteractive
```
johnspackman, line 65: You dont need the separate directory `myapp` within an outer `testdrive`, you can just use `--out`:

```
mkdir testdrive
cd testdrive
npm init -y
npm install @qooxdoo/compiler
npx qx create testdrive --noninteractive --out=.
npx qx package install qooxdoo/qxl.apiviewer
npx qx package install qooxdoo/qxl.widgetbrowser
npx qx serve
```

derrell:
Replaced by #652

PR #531, opened 2019-08-07
 fixes an issue when there is no compile.json at all (just compile.js)

johnspackman:


source/class/qx/tool/cli/commands/Package.js
```diff
@@ -172,12 +163,6 @@ qx.Class.define("qx.tool.cli.commands.Package", {
           qx.tool.compiler.Console.info(`>>> Saved library data to ${lockfileModel.getRelativeDataPath()}`);
         }
       }
-      if (compileConfigModel.isDirty()) {
```
cboulanger: Can this be safely deleted? Are all changes to compile.json manually safed?

johnspackman: yes, because Package no longer uses it - but no, because Install didn't have the save (see next commit)

PR #538, opened 2019-08-12
This turns off tagging beta releases as a "prerelease" when `qx publish`ing them

cboulanger:
If you want releases to be "prereleases" in a GitHub sense (not in a semver sense), use the `--prerelease` parameter


johnspackman:
thank you @cboulanger

PR #539, opened 2019-08-13
This re-adds the build version of the "qx serve" website

cboulanger:
It wasn't a good idea after all to gerate the built version on demand since it can get out of sync and we have no auto-update mechanism in place


cboulanger:
This problem was only showing when using a git checkout, which would not update the build files. NPM version did not have the problem.

PR #550, opened 2019-08-19
Fixes #549

### LINKED ISSUE #549: Get rid of "fs.promises API is experimental" warning
Unfortunately, the warning is not going away until node version 12, and it is not possible to set `--no-warnings` in the shebang of the `qx` executable because it is not supported by Travis and older Linux distributions.

Until then, we really need to get rid of this annoying warning by converting the `qx` executable into a wrapper script which calls node with `--no-warnings`.

cboulanger:
Fixes #549

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,0 +1,2 @@
+require("./index");
+new qx.tool.cli.Cli().run();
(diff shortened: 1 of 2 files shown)


johnspackman:
Hi,

Thank you for your email - please note that I am on annual leave, returning on September 9th.  
 If you have an emergency support issue, please describe your issue in detail in email and then send a text message to the mobile number on your invoice.

Regards
John Spackman

PR #552, opened 2019-08-20
Json to ast

johnspackman:
This removes dependency on the external https://github.com/johnspackman/json-to-ast repo that was linked via github URL; the original author declined to integrate the changes, and rather than create yet another npm module that must be maintained, the code is integrated into the `qx.tool.*` namespace.


source/class/qx/tool/cli/commands/Command.js
```diff
@@ -146,8 +145,8 @@ qx.Class.define("qx.tool.cli.commands.Command", {
     parseJsonFile : async function(filePath) {
       var data = await fs.readFileAsync(filePath, "utf8");
       try {
-        let compileAst = JsonToAst.parseToAst(data, {verbose: true});
-        let compileJson = JsonToAst.astToObject(compileAst);
+        let compileAst = qx.tool.utils.json.Parser.parseToAst(data, {verbose: true});
```
hkollmann: Should´nt this use qx.tool.utols.Json.parseJson?

cboulanger:
Will the `test` dir be part of the NPM package? If yes - could you include something to prevent that into your PR?

hkollmann:
You can add it to .npmignore!

Von meinem iPad gesendet

cboulanger:
Yes, `test` should be added to `.npmignore`...

source/class/qx/tool/cli/commands/Command.js
```diff
@@ -94,7 +94,7 @@ qx.Class.define("qx.tool.cli.commands.Command", {
       let qooxdooJsonPath = path.join(process.cwd(), qx.tool.config.Registry.config.fileName);
       let data = {};
       if (await fs.existsAsync(qooxdooJsonPath)) {
-        data = await this.parseJsonFile(qooxdooJsonPath);
+        Ωdata = await qx.tool.utils.Json.loadJsonAsync(qooxdooJsonPath);
```
cboulanger: There is an "Ω" before `data` - this is probably by accident, no?

PR #558, opened 2019-08-24
DO NOT MERGE. Proposal: Add a second qx executable for debugging

cboulanger:
This is just an example, hasn't been tested at all. 

At the moment, this would start the node process as it was before
without suppressing warnings. Later, this would give us the possibility to start node with run-time flags to do more elaborate debugging or test new node features.


hkollmann:
In qx-debug.cmd you have to call qx-debug

Von meinem iPad gesendet

johnspackman:
I think there's limited value in doing this - developers can find their own way around this problem, it's just annoying that there is an issue in the first place.  EG I've edited `qx` in my private master to be the old version.

Creating sub processes is IMHO messy and best avoided - off the top of my head, these are the problems it introduces:

* it crufts the process list
* the node process can no longer change it's process name
* signals do not work without even more code in the parent bash script to forward them on

Also, I dont know what's changed but I just switched back to node and 11 did this *without the warning*:

```
johns-mbp:~ john$ node -v
v11.15.0
johns-mbp:~ john$ cd dev/Qooxdoo/QxCompiler
johns-mbp:QxCompiler john$ cd test/testapp/
johns-mbp:testapp john$ rm -rf compiled
johns-mbp:testapp john$ cat `which qx`
#!/usr/bin/env node

require("./index");
new qx.tool.cli.Cli().run();
johns-mbp:testapp john$ qx compile
File source/index.html exists but is no longer used
[ ... snip ...]
```

johnspackman:
Ah, the warning is in node 10, sorry ...

cboulanger:
Ok, closing this in favor of #559. Can be reopened if needed.

PR #562, opened 2019-09-02
enable travis testing with windows and fix windows tests

hkollmann:


source/class/qx/tool/compiler/resources/Manager.js
```diff
@@ -293,23 +293,15 @@ qx.Class.define("qx.tool.compiler.resources.Manager", {
...
+      let library = this.findLibraryForResource(srcPath);
       if (!library) {
         this.warn("Cannot find library for " + srcPath);
         return null;
       }
 
+
       let resourceDir = path.join(library.getRootDir(), library.getResourcePath());
-      srcPath = path.relative(resourceDir, path.join(resourceDir, srcPath));
+      srcPath = path.relative(resourceDir, path.isAbsolute(srcPath)?srcPath:path.join(resourceDir, srcPath));
```
johnspackman: this will cause problems - just about everywhere, the asset's path is expected to be relative to the resource directory of the library; `asset.getSourceFilename()` will join the paths, and uses it to check whether the resource is in the theme or resource directory.

If srcPath is absolute it should be reduced to relative

hkollmann: The srcPath is relative to resourceDir. Problem occours if srcPath is absolute - then the join with resourceDir will add resourceDir to the absolute path

hkollmann:
Test are working now for Linux + Windows, node 10 + 12 + current

source/class/qx/tool/compiler/resources/Manager.js
johnspackman: oh, sorry you're right

johnspackman:
thank you @hkollmann

PR #563, opened 2019-09-04
for windows we need upath here

hkollmann:


cboulanger:

Issue #440, opened 2019-06-07
Confirm that exit status is non-zero on compile error


derrell:
For unattended operation, e.g., to compile the app from gitlab-ci, it is important that if the compiler detects an error, it exits with a non-zero status. According to @hkollmann it may already do this properly, but @johnspackman requested an issue to confirm that it does.

hkollmann:
Any idea how to test this?

derrell:
There are at least two categories of errors that should result in non-zero exit status:

- a compile error, e.g., invalid JavaScript
- valid JS, but includes a reference to a class that can not be resolved, e.g., a library is not available or a name is mistyped in an @require, e.g., a "link" or dependency error

I suspect there are additional, common categories of errors but identifying them probably requires better understand than I possess regarding the internals of this compiler.

cboulanger:
Shouldn't **all** errors result in non-zero exit status?

derrell:
**Yes**.

I took the question to be how to write some test cases that should cause errors to confirm that non-zero exit codes result.

cboulanger:
@derrell @johnspackman What is the status of this issue?

oetiker:
Just looked through the code and it is not immediately obvious how to tackle this.

```diff
--- a/source/class/qx/tool/compiler/makers/AppMaker.js
+++ b/source/class/qx/tool/compiler/makers/AppMaker.js
@@ -164,7 +164,7 @@ qx.Class.define("qx.tool.compiler.makers.AppMaker", {
             application.calcDependencies();
             if (application.getFatalCompileErrors()) {
               qx.tool.compiler.Console.print("qx.tool.compiler.maker.appFatalError", application.getName());
-              return undefined;
+              throw new Error("FatalError");
             }
 
             this.fireDataEvent("writingApplication", application);
```
seems a whee bit drastic and also breaks `--watch`

johnspackman:
where is that `throw new Error("FatalError")` ?  It's not in the current master https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/makers/AppMaker.js#L165-L167

oetiker:
> where is that `throw new Error("FatalError")` ? It's not in the current master https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/makers/AppMaker.js#L165-L167

that is just a little experiment I did :) to see if there was a quick fix

[...]

PR #565, opened 2019-09-05
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/440

johnspackman:
In the original request in https://github.com/qooxdoo/qooxdoo-compiler/issues/440#issuecomment-499993805, @derrell asked for things like missing class references to also trigger a non-zero exit status - this has kind of been done, but missing references are "just" unresolved symbols which are just warnings.  As they're warnings, they don't stop the class from being compiled, so even if a non-zero exit status was given, the next time you ran `qx compile` you'd get a zero exit status.

IMHO it's a pain anyway if warnings are only displayed if a class is compiled - ie if the class is not edited, you don't see the warning even if that warning could affect code you *are* editing.

As well as non-zero exit codes, this PR also displays warnings for classes that have not been compiled, and also the `--warn-as-error` argument will cause a non-zero exit code if there are any warnings.

Although this is not quite as exact as detecting a missing class, my thought is that as this may have the same net effect.


derrell:
That sounds like it should allow accomplishing the goal: ensure that any `qx compile`-detected error or warning stops further steps in a scripted build process. Thanks!

oetiker:
does this the also cause the compiler to set an exit value on syntax error ?

johnspackman:
yes

cboulanger:
@hkollmann yargs will translate "warn-es-error" to "warnAsError" implicitly (since hyphens cannot be used as member names unless in quotes and square brackets)

johnspackman:
@hkollmann also AIUI if you tell yargs the name is `warn-as-error` you can use `--warn-as-error` or `--warnAsError`, and reference it as `argv["warn-as-error"]` or `argv["warnAsError"]`.  But (I think) that if you tell yargs the name is `warnAsError` then only `--warnAsError` works and only `argv["warnAsError"]` in the code

johnspackman:
thanks for patching this @hkollmann 👍

PR #574, opened 2019-10-04
fixes a bug in parsing jsdoc

johnspackman:
fixes a bug in parsing jsdoc, where the last compiler hint won't be parsed properly if there is anything on the next line

### DIFF
--- a/source/class/qx/tool/cli/ConfigDb.js
+++ b/source/class/qx/tool/cli/ConfigDb.js
@@ -25,7 +25,7 @@ const path = require("path");
 qx.Class.define("qx.tool.cli.ConfigDb", {
   extend: qx.core.Object,
   
-  construct() {
+  construct: function() {
     this.base(arguments);
     this.__overrides = {};
   },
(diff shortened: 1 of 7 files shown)


johnspackman:
seems unlikely that's down to this PR, given the extent of the change.  Here's my output:

```
Johns-MacBook-Pro:test john$ cd ../../QxCompiler
Johns-MacBook-Pro:QxCompiler john$ ./qx package update
Downloading cache from GitHub ...
Found 287 releases of libraries.
Run 'qx package list' in the root dir of your project to see which versions of these libraries are compatible.
Johns-MacBook-Pro:QxCompiler john$
```

johnspackman:
This was weird - this syntax does not work in native code:

```
qx.Class.define("blah.Class", {
  construct() {
    this.base(arguments);
  }
});
```

you get the "is not a constructor" message when trying to `new blah.Class()`.  I've had this before, but do not understand how this slipped through before.  This PR is now updated to include fixes for ConfigDb and everywhere else there was the same signature.

The fix BTW is to use `construct: function() {}` instead of ES6 style `construct() {}`

johnspackman:
PS - for code which is compiled using `qx compile` the ES6 syntax works just fine...

PR #578, opened 2019-10-11
fix for https://github.com/qooxdoo/qooxdoo-compiler/issues/570

hkollmann:
add localstorage to globals


cboulanger:

PR #586, opened 2019-10-29
Fix warn as error

hkollmann:


johnspackman:
thank you @hkollmann !

PR #587, opened 2019-10-31
use of defaultNumberingSystem when loading locale

rommni:
In locale data provided by CLDR, there is for some language a field named "defaultNumberingSystem" in numbers, which give the numbering system that should be used by default.

Currently we are not using this and just take the first one like it's often in alphabetical order it's often the bad one. (for exemple fr locale use data of arabic numbering system instead of latin)

In this PR I check if this field is defined, and if it is i use the numberingSystem indicate in the field.

### DIFF
--- a/source/class/qx/tool/compiler/app/Cldr.js
+++ b/source/class/qx/tool/compiler/app/Cldr.js
@@ -245,8 +245,18 @@ qx.Class.define("qx.tool.compiler.app.Cldr", {
             cldr.cldr_time_format_short = find(timeFormatLength, "type", "short", getTimeFormatPattern);// "h:mm a";
           }
 
-          cldr.cldr_number_decimal_separator = getValue("ldml.numbers[0].symbols[0].decimal[0]");// ".";
-          cldr.cldr_number_group_separator = getValue("ldml.numbers[0].symbols[0].group[0]");// ",";
+          var numberingSystem = getText(get("ldml.numbers[0].defaultNumberingSystem[0]"));
+          
+          if (numberingSystem) {
+            find(get("ldml.numbers[0].symbols"), "numberSystem", numberingSystem, function(row) {
+              cldr.cldr_number_decimal_separator = row.decimal[0];
+              cldr.cldr_number_group_separator = row.group[0];
+            });
+          } else {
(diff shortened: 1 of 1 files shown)


cboulanger:
Could you add a test class for the test runner to ensure the correct result? Thank you.

rommni:
Of course I can do that, however just a question what do you think is the best place to test that? Should it be in test-compiler.js

cboulanger:
see my updated comment - the right place is probably `qx.test.locale.Locale`

rommni:
oh thank you was looking in the qooxdoo-compiler not directly in qooxdoo, so i have to do an other PR directly in qooxdoo repo? (sorry not really familiar to work in different but linked repo)

hkollmann:
This is compiler related - so we must test it in compiler. The old python toochain is not using defaultNumberingSystem. 
@cboulanger : What do you want to test?

cboulanger:
@rommni @hkollmann You're right. I didn't look carefully enough in which repo this PR was created! If this would break compiler builds, of course we cannot put the test in the qooxdoo framework repo (yet), altough eventually it should be part of the main test suite. 

@hkollmann We should have a test to make sure the change works as intended and that there won't be a regression in the future - shouldn't this be part of any code change that affects existing behaviour? However, this is a special situation in which the behavior depends on the toolchain.

cboulanger:
@rommni @hkollmann We don't have a place for these kind of tests yet, it seems. Tests in the compiler repo concern compilation behavior, not framework code behavior. The cleanest solution would be to add a test in the qooxdoo repo which runs it dependent on the compiler version in the environment variable `"qx.compilerVersion"`.

cboulanger:
@rommni @hkollmann But this can be added only after this change has been integrated and published.

@rommni  maybe you could locally add the test to `qx.test.locale.Locale`? Then we could integrate it later to the qooxdoo framework, using `if (!qx.core.Environment.get("qx.compilerVersion"){return}` to skip it in case of generator builds.

rommni:
@cboulanger of course i can look at this (not today but maybe during next week), maybe i could even create a qx.test.locale.Number like we have a Date one, test the value of group & decimal sign in different locale with and without the defaultNumberingSystem field, and test the number percent format which is the other information we get from locale related to number, we can then add more test if we decide in future to get more information from locale file.

johnspackman:
merging, thank you @rommni !

PR #589, opened 2019-11-01
NodeJS DOM Emulation

johnspackman:
Various parts of Qooxdoo depend on there being global objects for `window` and `document`, and also that these behave like their browser DOM equivalents to one degree or another.  In many circumstances, a very basic emulation is acceptable but it is possible to incorporate code which requires a much more complete implementation.  

The best (worst) example of this is `qx.bom.Selector` which does lots of feature-detection tests during class initialisation that break badly on NodeJS; this class is depended on by lots of other classes, so even if not used at runtime it causes an issue.  If you want to manipulate DOM-like structures on the server (eg you want to use Qooxdoo's upcoming JSX support ;) ) you may want to use qx.bom.Selector anyway.

This PR adds conditional support for JSDOM to be used *if it is available* - i.e. if you do not add JSDOM via `npm i jsdom` then the previous behaviour is unchanged.

### DIFF
--- a/source/class/qx/tool/cli/commands/Serve.js
+++ b/source/class/qx/tool/cli/commands/Serve.js
@@ -21,6 +21,7 @@ const process = require("process");
 const express = require("express");
 const http = require("http");
 const fs = qx.tool.utils.Promisify.fs;
+const chokidar = require("chokidar");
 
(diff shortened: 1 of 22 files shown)


johnspackman:
closinfg because the commit history is messed up

PR #598, opened 2019-12-04
Add support for setting application variables

voger:
I don't know if it is ready for merge yet. I opened this PR mostly to get some feedback about the approach I took. Manual testing worked. I plan to write some tests for it.

My main concerns: 

* I introduced a new option `--set-env` to set the applications environment variables. The idea is that since `--set` sets compiler options and the command `set` manages the DB then we wouldn't want to have generic environment variables available to all of our projects. This might introduce vulnerabilities. For example someone may store a security token there, not wanting to be available in every application. How would it be possible to exclude those? Also, the environment variables are set in `compile.json` and may be entered with a command line option. Would it make sense to introduce yet another place in the DB file to store environment variables?

* The parsing code I used in `MConfig.js` is almost identical with the parsing code from [`qx.tool..cli.commands.Command` line 63](https://github.com/voger/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/commands/Command.js#L63). Where would it be a good place to put a function to merge them somehow so it can be a bit more DRY?

Any thoughts?

### DIFF
--- a/source/class/qx/tool/cli/commands/MConfig.js
+++ b/source/class/qx/tool/cli/commands/MConfig.js
@@ -209,6 +209,7 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
         config.environment = {};
       }
 
+      qx.lang.Object.mergeWith(config.environment, parsedArgs.environment, true);
       if (!config.libraries) {
         config.libraries = [ "." ];
       }
@@ -251,6 +252,20 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
       if (argv.locale && argv.locale.length) {
         result.locales = argv.locale;
       }
+
+      if (argv["set-env"]) {
+        argv["set-env"].forEach(function(kv) {
+          var m = kv.match(/^([^=\s]+)(=(.+))?$/);
+          if (m) {
+            var key = m[1];
+            var value = m[3];
+            result.environment[key] = value;
+          } else {
+            throw new qx.tool.utils.Utils.UserError(`Failed to parse environment setting commandline option '--set-env ${kv}'`);
+          }
+        });
+      }
+
       return result;
     },
 
--- a/source/class/qx/tool/cli/commands/MConfig.js
+++ b/source/class/qx/tool/cli/commands/MConfig.js
@@ -256,13 +256,9 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
       if (argv["set-env"]) {
         argv["set-env"].forEach(function(kv) {
           var m = kv.match(/^([^=\s]+)(=(.+))?$/);
-          if (m) {
-            var key = m[1];
-            var value = m[3];
-            result.environment[key] = value;
-          } else {
-            throw new qx.tool.utils.Utils.UserError(`Failed to parse environment setting commandline option '--set-env ${kv}'`);
-          }
+          var key = m[1];
+          var value = m[3];
+          result.environment[key] = value;
         });
       }
 
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5798,6 +5798,11 @@
         }
       }
     },
+    "object-hash": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/object-hash/-/object-hash-2.0.1.tgz",
+      "integrity": "sha512-HgcGMooY4JC2PBt9sdUdJ6PMzpin+YtY3r/7wg0uTifP+HJWW8rammseSEHuyt0UeShI183UGssCJqm1bJR7QA=="
+    },
     "object-inspect": {
       "version": "1.4.1",
       "resolved": "https://registry.npmjs.org/object-inspect/-/object-inspect-1.4.1.tgz",
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "mkdirp": "^0.5.1",
     "node-fetch": "^1.7.3",
     "node-sass": "^4.12.0",
+    "object-hash": "^2.0.1",
     "ps-tree": "^1.2.0",
     "replace-in-file": "^4.1.2",
     "rimraf": "^2.6.2",
--- a/source/class/qx/tool/cli/Cli.js
+++ b/source/class/qx/tool/cli/Cli.js
@@ -45,12 +45,21 @@ qx.Class.define("qx.tool.cli.Cli", {
 
       let yargs = require("yargs").locale("en");
       yargs.option("set", {
-        describe: "sets an environment value",
+        describe: "sets an environment value for the compiler",
         nargs: 1,
         requiresArg: true,
         type: "string",
         array: true
       });
+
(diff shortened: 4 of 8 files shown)


johnspackman:
your approach looks fine to me; I was previously thinking about just having the one set option, shared for config and environment, but you make a very good point about security and actively *not* wanting some variables pushed into the environment.  

One change I would ask for though is that `--set-env` outputs a warning to the console about it's use - the reason is that the transpiler will use the environment variables to determine how to eliminate unused code; for example this code:

```
if (qx.core.Environment.get("qx.debug")) {
  doSomething();
} else {
  doSomethingElse();
}
```

That entire `if` statement will be deleted and replaced with either `doSomething()` or `doSomethingElse()` (this happens in every compile, whether source or build targets).  What the compiler does is to analyse the environment variables and see which ones differ between applications - if a value (such as `qx.debug`) is the same for every app, then every time the variable is referenced, the code will be edited on the fly to remove the unused part.  Only if the variable is different for different apps, then the whole `if` statement remains so that the `if` test is done at runtime.

So if you compile a source target with `--set-env qx.debug=false` and later do an incremental compilation (eg compiling only one file) but forget the `--set-env`, then that one file would be transpiled with debug code, even though the application as a whole has it turned off.  In practice that's possibly OK for most `qx.debug` cases, but some env variables will be more of an issue, and could lead to some very odd errors at runtime.

That's is why I quite like the idea of globally configured variables, they are hard to use inconsistently (although not impossible!).

_(PS I'm not saying that you need to find some perfect solution as part of this PR, I just wanted to take the opportunity to expand a little.  Food for thought etc :) )_

One solution to the security issue might be to have environment variables which are marked as private with underscores, eg a key called `qx.compiler._githubToken` that is never pushed to the code.  Or, because the compiler already records which environment variables are used by each class, it would be possible to only provide variables at runtime which are legitimately required.

Something the compiler needs to do anyway is to check for changes in the environment variables and if a change is detected then delete the transpiled directory and force a full compile (I think this is on the todo list already)

voger:
@johnspackman What if we store somewhere, in a file, a hash of the environment variables object for each build? Then, in the next build, we compare the new hash with the old hash. If they differ or if there is no old hash the compiler preforms a clean build. Otherwise it performs an incremental build. Would this be a good solution for the problem?

johnspackman:
yes, that would be fine; that hash could be stored in `db.json` by [Analyser.js](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/Analyser.js#L170).  A change in environment would mean that the entire database can be deleted and recreated,

voger:
I am searching through all the code but I can't find the proper place where the decision is made whether to do a full or incremental compile. 

I found this [spot here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/commands/Compile.js#L393) where the compiler cleans the whole output directory. Unfortunately at this point the analyser database is not open and we don't have access to the whole environment object, only the values from the command line.

I am also looking at the `qx.tool.compiler.makers.AppMaker` class and `qx.tool.compiler.targets.Target.generateApplication/2` method. Still cannot find how it decides whether to build the whole thing or only partially.

Any hints on where to look or how it works?

johnspackman:
Ah, yes - it's also called by `maker.checkCompileVersion()` [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/makers/Maker.js#L147-L162) via `AppMaker.make()` [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/makers/AppMaker.js#L133-L143) immediately before the database is opened.

I think that the `checkCompileVersion()` and `writeCompileVersion()` [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/makers/Maker.js#L143-L181) writes to a `version.txt` and `versions.json` partly to get around this chicken and egg issue, but that's a bit simplistic and do we really need yet another file?  (I wrote that code so I'm only criticising myself here 😆 )

The only reason that `Analyser.open()` is expensive is because it also calls `Analyser.initialScan()`, so I would suggest changing that so that `initialScan()` is called manually and `Analyser.open()` is called as soon as it is created eg [here](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/makers/Maker.js#L208); this means that in `AppMaker.make` you just read any environment or versioning information, and if you decide that a full rebuild is necessary you can call `maker.eraseOutputDir()` and some new method like `Analyser.eraseDatabase()` to reset the analyser.  When you're ready to go ahead and compile (full or incremental) you then call `initialScan()` and continue as before.

If you also chose to integrate the version information into the analyser's database (ie get rid of `version.txt` and `versions.json`) at the same time that would be cool too, but it's entirely up to you whether you do that.

voger:
What do you think?

johnspackman:
I can't remember why there was a versions.json, possibly there was a need to read it outside of the compiler.

I've approved the PR in the mean time but if you wanted to go ahead and move the version numbering into `db.json` as well that would be great!  I think that both `version.txt` and `versions.json` can be disposed of, in favour of somewhere in `db.json`

cboulanger:
If I remember correctly, @hkollmann added the `versions.json`, is that correct, @hkollmann, or am I mistaken?

hkollmann:
That's correct. We need a list of libs and there versions compiling to document the versions. Could go into db.json.

voger:
Give me few more days. I am working towards moving these files in `db.json`

hkollmann:
All time of the world 😄

voger:
* Removed `checkCompileVersion()` and `writeCompileVersion()` from `qx.tool.compiler.makers.Maker`
* Check for whether the environment has changed it done in `Analyser.isContextChanged`
* `Analyser.initialScan()` doesn't save the database as soon as it is finished. The database is saved in `AppMaker.make()` anyway.

source/class/qx/tool/cli/commands/MConfig.js
```diff
@@ -251,6 +252,20 @@ qx.Mixin.define("qx.tool.cli.commands.MConfig", {
       if (argv.locale && argv.locale.length) {
         result.locales = argv.locale;
       }
+
+      if (argv["set-env"]) {
+        argv["set-env"].forEach(function(kv) {
+          var m = kv.match(/^([^=\s]+)(=(.+))?$/);
```
cboulanger: The regex will not recognize `--set-env foo = bar` which might be sensible but this should be documented. It accepts `--set-env foo`, setting foo to be undefined - is there any use in this? Maybe setting a value of a key should be enforced?

voger: It works without spaces around the equal sign. i.e. `--set-env foo=bar`. It is the same regexp used for the `--set` option. Otherwise the shell or yargs would probably get confused and consider `foo`, `=` and `bar` as 3 different parameters.

cboulanger:
Thank you, @voger and happy holidays!

PR #600, opened 2019-12-13
Support for TC39 proposal plugins

johnspackman:
This adds support for additional, non-standard babel plugins - EG the `@babel/plugin-proposal-*` plug ins which add early proposed additions to javascript.

Not surprisingly this comes with caveats, first of which is that changes to the language may well require an upgrade to QxCompiler itself so that it can recognise the new language construct and act accordingly.  

In many cases, turning on a new plugin that QxCompiler does not support will just cause extra warnings, usually about unresolved symbols because (EG) QxCompiler did not understand a new syntax of declaring the variable in the first place.  But there may be other side effects where the plugin simply cannot work without an upgrade.

So, adding in babel plugins that extend the language is at your own risk, but should be useful.

This PR adds support for the [@babel/plugin-proposal-optional-chaining`](@babel/plugin-proposal-optional-chaining).

To enable this, you need to modify the top level `babel` property, eg:

```
  "babel": {
    "plugins": {
      "@babel/plugin-proposal-optional-chaining": true
    }
  },
```
Note that the old top level `babelOptions` property has been deprecated in favour of `babel.options`

### DIFF
--- a/source/resource/qx/tool/schema/compile-1-0-0.json
+++ b/source/resource/qx/tool/schema/compile-1-0-0.json
@@ -335,9 +335,25 @@
         "type": "string"
       }
     },
+    "babel": {
+      "type": "object",
+      "properties": {
+        "options": {
+          "type": "object",
+          "description": "Options given to @babel/preset-env. With this options the output type of babel can be defined. For details see here: <https://babeljs.io/docs/en/babel-preset-env#options>. They can be overridden per target."
+        },
+        "plugins": {
+          "type": "object",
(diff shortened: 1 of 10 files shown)


source/resource/qx/tool/schema/compile-1-0-0.json
```diff
@@ -335,9 +335,25 @@
...
+          "additionalProperties": {
+            "type": [ "boolean", "object" ]
+          }
+        }
+      }
+    },
     "babelOptions": {
       "type": "object",
-      "description": "Options given to @babel/preset-env. With this options the output type of babel can be defined. For details see here: <https://babeljs.io/docs/en/babel-preset-env#options>. They can be overridden per target."
+      "description": "** DEPRECATED - See babel.options instead"
```
hkollmann: This is only valid for the top level babelOptions element. So you should copy the old description to the target/babelOptions tag.

If we update the schema to https://json-schema.org/draft/2019-09/schema we could use the deprecated keyword.

See https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3

PR #607, opened 2019-12-19
move to node-sass 4.13

hkollmann:
and fix security audits.

We should close #603 and use this instead - it's updating the npm-shrinkwrap.json too


johnspackman:
Thank you @hkollmann

PR #609, opened 2019-12-25
Override default environment

voger:
So far the `--set-env` command line option assigns environment options but doesn't override the options set by the individual target. This PR fixes that. 

Environment settings coming from command line must override the settings set in `compile.json`. This PR merges the incoming values from the argv with those of the target's that is to be compiled right now. If that target doesn't have an `environment` object, one is created and assigned to it. 

The reason for not using the global environment when target's environment doesn't exist is that the target's environment takes precedence over the global environment.


johnspackman:
thank you @voger

PR #610, opened 2020-01-05
add qx test

hkollmann:
This adds qx test to the compiler.

To use qx test you need to add a test contrib to your app, e.g. testtapper.

Idea:
qx serve fires a new event after server start. qx test listen for this event and fires an event runTests. 
After this the process is finished.

The test contrib has to register a event handler to runTests in his compile.js file.
This event handler has to return a promise which is resolved after all tests are done.

See testtapper contrib for an example.

### DIFF
--- a/source/class/qx/tool/cli/Cli.js
+++ b/source/class/qx/tool/cli/Cli.js
@@ -63,6 +63,7 @@ qx.Class.define("qx.tool.cli.Cli", {
           "Create",
           "Lint",
           "Run",
+          "Test",
           "Serve"
         ],
         "qx.tool.cli.commands");
--- a/source/class/qx/tool/cli/api/LibraryApi.js
+++ b/source/class/qx/tool/cli/api/LibraryApi.js
@@ -47,7 +47,40 @@ qx.Class.define("qx.tool.cli.api.LibraryApi", {
      */
     async afterLibrariesLoaded() {
       // Nothing
-    }
+    },
+  
+    /**
+     * 
+     * helper to load an npm module. Check if it can be loaded before
+     * If not install the module with 'npm install --no-save --no-package-lock' to the current library
+     * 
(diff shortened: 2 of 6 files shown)


cboulanger:
So the idea is that this class works like `qx serve`, with the only addition that it fires an event which the testing package can react to?

hkollmann:
it starts serve, run all tests and terminate the process after this.

hkollmann:
` process: async function() { 
      this.addListener("started", async () =>  { 
        let result = {errorCode: 0}; 
        let res = this.fireDataEvent("runTests", result); 
        res.then(() => { 
          process.exit(result.errorCode);           
        }); 
     }); 
      await this.base(arguments);`

cboulanger:
But what I mean is that it doesn't really run any tests by itself - it just signals to a test runner that *might* exist that it should start the tests, no? So if there is no test runner, nothing happens. Is there a way for the command know if a test runner has been loaded at all, and throw an error if not?

hkollmann:
Can we check if there are registered listeners to an event? If not no test runner is registered.

hkollmann:
Found it: http://www.qooxdoo.org/devel/api/#qx.core.MEvent~hasListener!method_public, hasListener

cboulanger:
Yes, that seems to be an pragmatic and efficient way - I was thinking of checking the installed packages, but this is much easier and completely sufficient!

cboulanger:
So you can check for registered listeners and inform the user that s/he has to to install a test runner if no such listener has been found.

cboulanger:
I was looking at the testtapper package but could not find the place where it listens to the `runTests` event - do you have merged your changes there yet?

cboulanger:
How does `qx test` relate to the "qxl.testrunner" package?

cboulanger:

PR #615, opened 2020-02-26
add beforeStart event 

hkollmann:
- with this it's possible to add own extensions to express module
rename started event to afterStart for naming consistence


hkollmann:
Only if you have used the started event yet - Only known is test cli module

PR #617, opened 2020-02-27
fix startpage in serve

hkollmann:
- startpage should'nt be shown if there is an default app.
   - startpage should be shown if there is no app too.


johnspackman:
thank you @hkollmann !

PR #621, opened 2020-03-05
Issue585

johnspackman:
Fixes ever-growing `db.json` bug


johnspackman:
close due to git history

Issue #619, opened 2020-03-04
Translations with msgids containing escaped quotes cause invalid entries in po files


rad-pat:
When generating po files for our application I noticed some entries similar to:

It seems that the generation process does not like strings with escaped double quotes and will cause generation of invalid entries upon recompilation.

Steps to reproduce

Create a minimal app.
Wrap a string in `this.tr()` and add some escaped quotes e.g.
```
var button1 = new qx.ui.form.Button(this.tr("Click me \"here\""), "app8/test.png");
```
Compile the app once with `-u true`, see the correct entry generated in en.po:
```
#: testapp/Application.js
msgid "Click me \"here\""
msgstr "
```
Compile the app again with `-u true`, see the original entry marked unused and an additional invalid entry generated:
```
#. NO LONGER USED
msgid "Click me \"here\""
msgstr ""

#: testapp/Application.js
msgid "Click me \\"here\\""
msgstr ""
```

Expected result:
Only one valid entry to be generated

Actual result:
Repeated compilation adds more and more invalid entries to the po files

Tested with qx compiler 1.0.0-beta.20200228-1557, nodeJS v10.16.3 on Ubuntu 18.04.4

johnspackman:
@rad-pat @oetiker Please can you retest when https://github.com/qooxdoo/qooxdoo-compiler/pull/622 is merged?

rad-pat:
@johnspackman, the problems appear to be fixed as far as I can see so I'll close the issue if @oetiker is satisfied too.

PR #622, opened 2020-03-05
Fixes ever growing db.json and double backslashes in translation strings

johnspackman:
Fixes ever growing db.json


johnspackman:
I think that this may also fix https://github.com/qooxdoo/qooxdoo-compiler/issues/619

oetiker:
@johnspackman if there is a `\n` or `\"` in a msgstr in a `.po` file they show up as is coming out of this.tr ... the old generator translated them properly ... the rules for this are described at the end of https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html

johnspackman:
@oetiker I can't reproduce this, i've added unit tests in https://github.com/qooxdoo/qooxdoo-compiler/pull/623 and it seems to be working, please can you compare to your test case?

Issue #374, opened 2019-04-25
updating a contrib with `provides` key kills existing entry


oetiker:
In qxl.testtapper/Manifest.json there is a `provides` key which gets integrated into the compile.json of the application using the contrib ... on `upgrade` the existing entry gets overridden ... this is not ideal ...

cboulanger:
Hm, that should not happen. Dependencies of libraries should only go into `contrib.json` (which will be called differently soon). Can you put together a set of commands that will show the problem reproducibly?

hkollmann:
Only partly true. If a contrib provides an application it will be copied into the compile.json. I think @oetiker means that this should not be overridden with upgrade

oetiker:
yes ... this happens with qxl.testtapper and qxl.apiviewer

although the whole aproach might be a bit limited anyway since the contrib can only provide a single application in this way

cboulanger:
Ah, sorry, I misunderstood. This is about `provides`. not `requires`. But still I don't fully get it. Which upgrade do you mean - `qx contrib upgrade`? Can you give an example of what you do that produces the undesired outcome?

oetiker:
qx contrib upgrade (I think) or also qx contrib install new-version ... 

when I install the qxl.testtapper then the data from Manifest.json/provides gets added into the complie.json of my application ... replacing any customized version of that code from a previous install .

cboulanger:
@johnspackman Can you quickly point me to where in the code the copying of applicaton data from the manifest to compile.json happens`? This should be easy to fix. Or can you do it when you're working on bugs anyways?

johnspackman:
@cboulanger I'll take a look

hkollmann:
@cboulanger : qx.tool.cli.commands.package.Install.__installApplication

johnspackman:
thanks, i've got it

cboulanger:
@hkollmann of course, it’s in the club code, not in the compiler code, sorry @johnspackman! But if you can have a look nonetheless, it would be nice!

johnspackman:
@oetiker is this still happening?  I could not reproduce it

oetiker:
to reproduce

install qxl.testtapper 0.3.x

then edit the compile.json to fit your needs

the install version 0.4.0

in my setup this resets what is in compile.json

hkollmann:
Done with https://github.com/qooxdoo/qooxdoo-compiler/pull/625

PR #625, opened 2020-03-08
fix for #374

hkollmann:


hkollmann:
I just removed code which do nothing. Save flag is not been set so saving will not happen.
And the copy is done only if the app is not in the config file (!app)

PR #632, opened 2020-03-20
adds markdown support to comment parsing

hkollmann:
fix issue https://github.com/qooxdoo/qxl.apiviewer/issues/26: HTML markup and code sections in source code docblocks not correctly converted.


test/test-jsdoc.js
```diff
@@ -1,6 +1,93 @@
...
+* representing the remote resource is instantiated and actions are invoked on this resource.
+* A resource with its actions can be configured declaratively or programmatically.
+*
+* There is also {@link qx.io.rest.Resource} which uses {@link Resource} under the hood.
+* The main differences between them are:
+*
+* * The event object available in the listeners (e.g. @success()@, @getSuccess()@ and @getError()@) is
+*   a native JavaScript object instead of a qooxdoo object ({@link qx.event.type.Rest}):
+* ** See {@link qx.io.rest.Resource} vs. {@link Resource}
+* ** @event.getId()@ => @event.id@
```
cboulanger: What does the encapsulation in `@` do?

test/test-jsdoc.js
```diff
@@ -1,6 +1,93 @@
...
+	let text = `
+* *strong*
+* __emphasis__
+* @light@              -> <code>light</code> 
+* @light@  @light@            -> <code>light</code> <code>light</code>
+* {@link Resource}     -> link?
+* @ignore(qx.*)
+`  
+  var test = qx.tool.compiler.jsdoc.Parser.parseComment(text);
+  console.log(test["@description"][0].body);
```
cboulanger: This doesn't really test anything, does it? Maybe you could save the output in a file and use the content as the "expected" value?

test/test-jsdoc.js
hkollmann: @xxx@ found in some __index__.js covers the text into a <code>xxx</section>.

test/test-jsdoc.js
hkollmann: Next time

test/test-jsdoc.js
cboulanger: Ok, like backticks in MarkDown or `<code>...</code>`? We should replace this eventually, but that's beyond the scope of your PR.

hkollmann: You can use backticks now. It's an old - undocumented - qooxdoo feature.
Some others - like __xx__ are documented.
Just found it in your failure examples. 

http://www.qooxdoo.org/current/api/#qx.bom.rest :  The higlighted parts.

PR #634, opened 2020-03-21
add error handling around markdown conversion

hkollmann:


hkollmann:
This is a problem of deleting the docComment code.
I will revert this and add code for docComment in the apiviewer

PR #636, opened 2020-03-24
Fix param comment

hkollmann:
old fix is not working for each case, see qx.dom.Document.getWitdh

### DIFF
--- a/source/class/qx/tool/compiler/jsdoc/Parser.js
+++ b/source/class/qx/tool/compiler/jsdoc/Parser.js
@@ -92,8 +92,9 @@ qx.Class.define("qx.tool.compiler.jsdoc.Parser", {
             qx.tool.compiler.Console.error(`Markdown conversion error: "${e.message}" found in \n${cmd.body.trim()}`);
           }
         } else {
+          cmd.body = cmd.body.trim();
           // Strip trailing single line comment
-          var m = cmd.body.match(/(^.*)(\/\/.*)$/m);
+          var m = cmd.body.match(/(^.*)(\/\/.*)$/);
           if (m) {
             cmd.body = m[1].trimRight();
             cmd.docComment = m[2];


hkollmann:
If we strip https://files.gitter.im/cboulanger/Mkf3/Bildschirmfoto-2020-03-24-um-13.28.31.png will not work any longer.
This is not valid jsdoc but valid qooxdoo style and it's used in more then one place.
We should'nt break BC here.

cboulanger:
But then we should change the RegEx to not match URLs, maybe with a negative lookbehind:`(?<!https|http):\/\/(.+)$`.

cboulanger:
And I think we should mark it deprecated and remove those comments for v7.

hkollmann:
I just changed the apiviewer to use docComment so that the http:// comments will be shown corrrect

cboulanger:
That sounds like quite hackish ... wouldn't it be better to just change the Regular expression?

hkollmann:
I do not know where this syntax is used in the whole framework. This is the reason why i do not like to deprecate it now because we will get a lot of deprecation messages during compile the framework

cboulanger:
Ok, forget the deprecation. But I suggest to update the regular expression and revert the APIViewer change that works around the wrong match for URLs. So we leave everything as it is and just "save" the URLs from being mininterpreted as a comment.

hkollmann:
What‘s about doing this code cleanup after releasing?
With the current solution no doc comment is lost.

cboulanger:
I don't insist, I just think it would be less work to fix it now, because all you need to change is the Regex - it would not lose any doc comment, it would just not falsely match URLs.

johnspackman:
Originally, I envisaged each command as a single line, so `@ignore(someword) // blah` is easily split into 3 parts: `{ name: "ignore", body: "someword", docComment: "blah" }`.

The underlying issue is that commands can actually be split over multiple lines, and if you're going to parse out the comments the dilemma is that they are positional, ie if you are going to preserve them, the comments are useless unless they remain next to the thing that they are commenting, so you have to leave them intact.  OTOH if you leave them in, then the parser hasn't done a very helpful job, and every single bit of code that implements a JSDOC comment has to parse out the comments.

IMHO this is a simple parser which is used to just get a list of `@xxx` commands and their parameters, and should be stripping out and discarding the single line comments.

EG this:

```
/*
 * @ignore(abc,
 *    def,
 *    ghi)
 */
```

is legitimate and becomes `{ name: "ignore", body: "abc,\ndef,\nghi" }`.

Single line comments are allowed, so this should have the same result:

```
/*
 * @ignore(abc, // abc comment
 *    def, //def comment
 *    ghi)
 */
```

(OK, with `@ignore` you could use 3 `@ignore` commands, but imagine something a little more complex with positional parameters and it's valid to want to comment)

URLs are valid:

```
/*
 * @ignore(stuff) // comment about ignore stuff
 * http://abc.com // comment about url
 * http://dev.com 
 */
 ```

Becomes `{ name: "ignore", body: "stuff" }`.

I'm preparing a PR with this in

hkollmann:
https://github.com/qooxdoo/qooxdoo-compiler/pull/637 is the better solution

PR #642, opened 2020-04-03
Add dynamic `qx` commands

johnspackman:
This PR refactors the processing of command line arguments and `compile.js[on]` so that the compiler API can be used to add commands to the `qx` tool chain - so for example, this allows packages that have been added by `qx package install` to extend the command line of `qx`, but only when running `qx` within that project directory.

For example, this would allow the `qxl.cms` package to add commands like `qx cms create-page`, or the API viewer to only reanalyse classes on demand when the user runs `qx api-viewer reanalyse`

The only BC issue is that you used to be able to specify the filename for `compile.js[on]` on the command line to `qx compile` - however, with this PR then *every* invocation of `qx` needs to be able to load the file, and therefore there needs to be standard command line switch to override it (we cannot rely on the position being immediately after the command keyword any more)


cboulanger:
Great stuff, which will be VERY useful for our package infrastructure!

I didn't see an implementation of the command line switch to determine the compile.json file, though ?

johnspackman:
I had to quickly test that :)

The rule is in `qx.tool.cli.Cli. __createYargs` https://github.com/qooxdoo/qooxdoo-compiler/blob/d29e6af0f05ebe6cde6be45e0f3db5b4f7411501/source/class/qx/tool/cli/Cli.js#L88-L92 and it's processed https://github.com/qooxdoo/qooxdoo-compiler/blob/d29e6af0f05ebe6cde6be45e0f3db5b4f7411501/source/class/qx/tool/cli/Cli.js#L207

hkollmann:
Do you have an example how to use this new feature?

johnspackman:
> Do you have an example how to use this new feature?

Yes - https://github.com/qooxdoo/qooxdoo-compiler/blob/master/test/testlib/compile.js#L26-L42 adds a new command called "testlib", which is a library used by https://github.com/qooxdoo/qooxdoo-compiler/tree/master/test/testapp

so you get this:

```
Johns-MacBook-Pro:QxCompiler john$ cd test/testapp
Johns-MacBook-Pro:testapp john$ qx testlib "Hello World" -t=6
The commmand testlib; message=Hello World, type=6
Johns-MacBook-Pro:testapp john$
```

cboulanger:
John, this will be very useful! For my JSON-RPC incubator, I immediately thought of a command that would generate client-side stubs for JSON-RPC server-side methods  so they could be used seamlessly on the client. 

Can you add a bit of documentation on this to our docs?

PR #647, opened 2020-04-09
Fix command line conflicts and npm package updates

johnspackman:
Fixes two bugs in command line processing, first it changes some of the aliases for command line switches to avoid conflicts with the new global `--config-file`, and also fixes a bug where positional config file argument to `qx compile` was accepted but ignored.  These two bugs caused the qooxdoo server deployment to be broken.

Also this includes updates to all npm packages (minor versions), and also updates to latest octokit/rest to get rid of that warning `[before-after-hook]: "Hook()" repurposing warning` which was breaking tests.


cboulanger:
Thanks for that octokit upgrade - the stupid warning has been bugging me and I had no idea where it came from.

PR #650, opened 2020-04-09
Fix broken "qx pkg publish" and qooxdoo.json support

cboulanger:


johnspackman:
thank you @cboulanger

PR #651, opened 2020-04-13
fixes watch for assets:

hkollmann:
- assets are not recognized from watch task.
   - assets are recognized only once


johnspackman:
thank you @hkollmann

Issue #604, opened 2019-12-17
single file in builds


oetiker:
when running the compiler in build mode it should only output a single file not multiple js file like boot.js, part-0.js, polyfill.js and resources.js

Issue #296, opened 2019-02-01
add --deploy-to path/to/deploydir


ecc7220:
The 5.0.2 version builds only one javascript and index.html plus resources. The new qx compiler puts plenty of other files into the build folder. They are not all needed for deployment and could cause a possible security issue, if installed as is.

johnspackman:
There are a few files which have crept into the build that I don't understand why they're there - `app.png` (which is the Qooxdoo logo) and `js_256x256.png` (which is some kind of javascript logo).

Of the rest of the files the only ones which are not required are `build-data.json` and `*.map`; the `.map` files are part of the target and need to be there, although I agree that they would normally be excluded from your deployment process.  

Potentially we could add a `--deploy-to <directory>` option to `qx compile` which would copy the essential files to a deployment directory?

ecc7220:
I agree, this would a good solution.

sqville:
I agree as well

derrell:
To add a bit more detail about this... The first application listed in `compile.json` is now the one whose `index.html` is copied to the root. Other necessary files, however, are not copied to the root, e.g., `favicon.png`, which means that if the app is accessed as `/myApp/index.html` the favicon shows on the tab, but if it's accessed as `/index.html`, the facicon is not shown on the tab.

oetiker:
when building multiple apps it would be great to have a way to specify a separate index.html or rather a separate boot directory for it ...

one option would be to structure boot the same as the resources directory in that it has subdirectories per app and if the app is missing then it gets the defaul settings ... 

whatever is in the toplevel boot would end up in the root of the *-output tree

cboulanger:
Updated title.

johnspackman:
For first compile of a source target, how about we output a message to the console in big friendly letters saying that the build version will be much smaller?

cboulanger:
Good idea. This will reach people who might not have read the info in the getting-started-guide.

An alternative to a deployment switch would be to output the paths to the directories containing the deployable code when using a "Build" target.

cboulanger:
It has just been brought up on the main chat that source maps leak information on paths on the development machine. They should be excluded from the deployed source code bundle. This reinforces the idea that we should have a dedicated command to export deployable code. This would avoid that we have to document such internal stuff.

cboulanger:
[...]

Issue #295, opened 2019-02-01
Transpiled build code size very long compared to version 5.0.2


ecc7220:
Compared javascript files from 5.0.2 project sums up to 1.1 MB in my case (only one file). Using 6.0.0-alpha-20181226-a3c2b0b and compiler 0.2.54 I get just for a few classes ported so far already 1.5 MB of file size, counting only part-0.js other files boot.js, resources.js and polyfill.js not included.

cboulanger:
@johnspackman Is that still the case and is there anything that can be done about this, and if yes, would it happen before the release of v6.0.0?

paragasu:
In my case the previous 5.0.2 generate about 2.1mb of js file. Ported into 6.0.0-alpha using qx compiler v0.3.0-alpha.20190430-1926 generate about 87mb files. 

(There is unoptimize resource directory full with icons svg files totaling about 21mb)

johnspackman:
@paragasu Resources are transpiled as well as source code because some resources are created during the compilation process (eg image combine/split); this means that if you have a lot of resources, the output directory will be large, but OTOH the compiler will not modify the source code or resources, and means that the output directory is self contained (which greatly simplifies generation and hosting).

However, I would expect that only those resources which are required by the application are copied into the output directory; the compiler knows which resources you need by looking for `@asset` directives in your code, so for example if your application uses the `myapp` namespace and you have `@asset(myapp/*`) that will include *all* resources in your output directory.

In the output directory, the only things you need for deployment are the `resource` and `myapp` (or whatever you application is called) directories - everything else are temporary working files for the compiler.  

In 5.x and earlier, the generator would create a cache directory which was similarly big (actually it could be a *lot* bigger than 80Mb), and folders like `transpiled` have a similar purpose - without those temporary files, compiling would be very slow.

AFAIAA, there is an increase in .js code size as reported by @ecc7220, but that's 1.1Mb => 1.5Mb, which is nowhere near the increase of 2.1Mb -> 66Mb.  Please can you take a look and confirm where the extra disk usage is coming from?

cboulanger:
The problem will go away if we have #296 implemented. As @johnspackman explained, the current directory structure makes visible what was before hidden by the generator, and makes use of more advanced javascript syntax feature (via Babel) possible - which the generator was unable to do.

hkollmann:
closed with https://github.com/qooxdoo/qooxdoo-compiler/pull/654

PR #654, opened 2020-04-20
Add `qx deploy`, i18n-as-parts, and improved target output

johnspackman:
This is a substantial reworking of the compilation process in `qx.tool.compiler.target`, which is to replace the POJO data structures that were undocumented and based on the generator's templates.  This PR builds proper OO data structures around the various components and makes it possible to reconfigure what source code is in what files as the target figures out what it needs to do, and then at the very end stream it all out to disk.

The changes are used to ensure the minimum number of files output, which for a build target will normally be just `boot.js` file (unless you have parts).

There are likely to be some compatibility issues (ie bugs, not BC problems) but as this is a refactoring as opposed to a complete rewrite from scratch they should be fairly straightforward  to resolve.

There is one change which might affect users, which is that the name of the script in the application's output directory is always `boot.js` - node applications used to be given the application name, but this is not done any more.

The other addition is that there is now `qx deploy`, and the `qx compile` will warn the user the first time they compile that `qx deploy` will produce a cut down version for them

fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/604
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/296
see also https://github.com/qooxdoo/qooxdoo-compiler/issues/295


source/class/qx/tool/cli/commands/Deploy.js
```diff
@@ -0,0 +1,158 @@
...
+      
+      if (argv.sourceMaps === null) {
+        qx.tool.compiler.Console.print("qx.tool.cli.deploy.sourceMapsNotSpecified");
+      }
+      
+      if (!argv.clean) {
+        qx.tool.compiler.Console.print("qx.tool.cli.deploy.notClean");
+      }
+      
+      config.targetType = "build";
```
cboulanger, line 84: This should probably be made configurable (later) so that one can supply the name of a custom build target

source/class/qx/tool/cli/commands/Deploy.js
```diff
@@ -0,0 +1,158 @@
...
+        describe: "deploys qooxdoo application(s)",
+        builder: {
+          "out": {
+            describe: "Output directory for the deployment",
+            demandOption: true
+          },
+          "app-name": {
+            describe: "The name of the application to deploy (default is all apps)",
+            nargs: 1,
+            type: "string"
```
cboulanger, line 46: Maybe a comma-separated list of app names would make sense. But one can also run `qx deploy --app-name={app1,app2}` in a shell.

source/class/qx/tool/cli/commands/Deploy.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+qx.Class.define("qx.tool.cli.commands.Deploy", {
+  extend: qx.tool.cli.commands.Command,
+  statics: {
+    getYargsCommand: function() {
+      return {
+        command: "deploy [options]",
+        describe: "deploys qooxdoo application(s)",
+        builder: {
+          "out": {
```
cboulanger, line 39: shorthand aliases would be good (but can be added later)

PR #657, opened 2020-04-22
Fix Node v8 and missing resources in `qx deploy`

johnspackman:
makes compiler work on node v8 again;
fixes regression in resources output;

### DIFF
--- a/test/testTargetsApp/Manifest.json
+++ b/test/testTargetsApp/Manifest.json
@@ -16,6 +16,12 @@
     "resource": "source/resource",
     "translation": "source/translation"
   },
+  "externalResources": {
+    "script": [ 
+        "https://localhost:2000/IDSCloudApp/IDSCloudApp:ext-js/socketcluster/socketcluster-client.min.js",
+        "http://localhost/compiled/source/cv/cv:libs/EventRecorder.js" 
+    ]
+  },
   "requires": {
     "@qooxdoo/framework": "^6.0.0-beta",
     "@qooxdoo/compiler": "^1.0.0-beta",
--- a/test/testTargetsApp/Manifest.json
+++ b/test/testTargetsApp/Manifest.json
@@ -18,8 +18,8 @@
   },
   "externalResources": {
     "script": [ 
-        "https://localhost:2000/IDSCloudApp/IDSCloudApp:ext-js/socketcluster/socketcluster-client.min.js",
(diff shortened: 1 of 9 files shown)


test/testTargetsApp/compile.json
```diff
@@ -10,7 +10,8 @@
 		},
 		{
 			"type": "build",
-			"outputPath": "compiled/build"
+			"outputPath": "compiled/build",
+            "minify": "off"
```
oetiker, line 14:
```suggestion
                          "minify": "off"
```

test/testTargetsApp/Manifest.json
```diff
@@ -16,6 +16,12 @@
     "resource": "source/resource",
     "translation": "source/translation"
   },
+  "externalResources": {
+    "script": [ 
+        "https://localhost:2000/IDSCloudApp/IDSCloudApp:ext-js/socketcluster/socketcluster-client.min.js",
+        "http://localhost/compiled/source/cv/cv:libs/EventRecorder.js"
```
cboulanger, line 22: What's the deal with the colon again? I have just looked into the docs but couldn't find anything on that - was that some sort of micro-protocol? If yes, we should document it, otherwise it looks like a bug.

johnspackman: no - i misunderstood the error, I though that was the actual URL, ie a directory name with a colon in it.

test/testTargetsApp/Manifest.json
```diff
@@ -18,8 +18,8 @@
   },
   "externalResources": {
     "script": [ 
-        "https://localhost:2000/IDSCloudApp/IDSCloudApp:ext-js/socketcluster/socketcluster-client.min.js",
-        "http://localhost/compiled/source/cv/cv:libs/EventRecorder.js" 
+      "https://localhost:2000/IDSCloudApp/IDSCloudApp:ext-js/socketcluster/socketcluster-client.min.js",
```
cboulanger, line 21: So here it is a directory with a colon in it, for test purposes?

johnspackman: yes because colons are allowed.  Library prefixes are no longer used

cboulanger: Ok, thank you!

cboulanger:
thank you @johnspackman !

Issue #659, opened 2020-04-22
qx compiler new version, windows


hkollmann:
Problems compiling with Windows, source target:

part from boot.js:

  packages : {
  "0": {
    "uris": [
      "package-0.js",
      "polyfill.js"
    ]
  },
  "1": {
    "uris": [
      "package-1.js"
    ]
  },
  "2": {
    "uris": [
      "..\\transpiled\\vzacore\\Configuration.js"
    ]
  },
  "3": {
    "uris": [
      "package-3.js"
    ]
  },

Error message: Can not load script….

hkollmann:
Done with https://github.com/qooxdoo/qooxdoo-compiler/pull/660

Issue #658, opened 2020-04-22
qx deploy: root index.html is not copied


hkollmann:
The root index.html as start page is not copied during qx compile

hkollmann:
done with https://github.com/qooxdoo/qooxdoo-compiler/pull/660

PR #660, opened 2020-04-24
fix for #659: qx compiler new version, windows

hkollmann:
fix for #658: qx deploy: root index.html is not copied


cboulanger:
thank you @hkollmann !

PR #661, opened 2020-04-26
FOR DISCUSSION ONLY:

valette:
load scripts using a fixed cache value defined as the compile time to solve caching issues. This PR only works with the following script added to index.html:
<script>
window.GLOBALCOMPILETIME = ${compileTime};
</script>

### DIFF
--- a/source/class/qx/tool/compiler/targets/Target.js
+++ b/source/class/qx/tool/compiler/targets/Target.js
@@ -670,6 +670,7 @@ qx.Class.define("qx.tool.compiler.targets.Target", {
 
       let pathToTarget = path.relative(path.join(t.getOutputDir(), t.getProjectDir(application)), t.getOutputDir()) + "/";
       let TEMPLATE_VARS = {
+        "compileTime": Date.now(),
         "resourcePath": pathToTarget + "resource/",
         "targetPath": pathToTarget,
         "appPath": "",
@@ -740,6 +741,7 @@ qx.Class.define("qx.tool.compiler.targets.Target", {
       if (application.getWriteIndexHtmlToRoot()) {
         pathToTarget = "";
         TEMPLATE_VARS = {
+          "compileTime": Date.now(),
           "resourcePath": "resource/",
           "targetPath": "",
           "appPath": t.getProjectDir(application) + "/",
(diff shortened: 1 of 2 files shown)


cboulanger:
Hi, can you describe in more detail what problem this PR is trying to solve? Maybe there is a different way that does not require to add additional code.

valette:
Hi,
this PR is for discussion only, I am facing issues with the caching mecanism on a server. Currently, there seems to have 2 ways of loading scripts with qooxdoo:
- loading scripts without dealing with cache
- loading scripts with cache busting i.e. adding ?nocache=Math.random()

but there is a third possible way : adding ?nocache=FIXEDVALUE where FIXEDVALUE would be defined at compile time.
With this approach, each new compilation would skip caching on the browser side, but once the script is loaded, a browser refresh would use the cached file as FIXEDVALUE did not change.

valette:
The patch is really dirty, I just wrote it to solve my problem quickly, and it would need more work if this approach was to be included in qooxdoo...

valette:
I forgot to add that the script loading part in index.html would become:
```
  <script type="text/javascript" src="${appPath}bundle.js?nocache=${compileTime}"></script>
  ${preBootJs}
  <script type="text/javascript" src="${appPath}boot.js?nocache=${compileTime}"></script>
```

johnspackman:
How about another solution - the scripts output into the boot.js file should have a timestamp which is the date that the file was last modified.  This would only happen if `addNoCacheParam` is true, and would limit the cache-busting to the exact minimum required.

If you wanted to code it that would be really helpful!

I've noticed that the addNoCacheParam has not been implemented in the compiler, so there are a few steps:
* Add a command line switch (in [Compile.js](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/commands/Compile.js)) and entry in `compile.js` ([schema](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/resource/qx/tool/schema/compile-1-0-0.json)) and then update the [global config](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/Cli.js#L419) to suit.
* Configure the target in [Compile.js](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/cli/commands/Compile.js#L665)
* Add a property `addNoCacheParam` to [Target.js](https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/targets/Target.js)
* The list of URIs is output here https://github.com/qooxdoo/qooxdoo-compiler/blob/master/source/class/qx/tool/compiler/targets/meta/Package.js#L245-L248 - change that so that each URI has the file's last modified date/time

valette:
Thank you for all the pointers, I will do that!

Issue #662, opened 2020-04-26
When a class file has Windows-style line endings (CR/LF) the @asset annotation does not always work.


andriy-s:
When a class file has Windows-style line endings (CR/LF) the `@asset` annotation does not always work.

It can easily be reproduced with the following three steps: (1) Create a new app (using `qx create`), (2) convert line-endings in the `Application.js` file to DOS/Windows style (using `unix2dos` or any other means), (3) compile. Result: the application resources are not copied (only qx resources are), the path to test.png is not adjusted, when the app is opened in the browser the “Click me” button is missing the icon. Is this a bug or an intended behavior? The old generator worked fine with either style of line endings.

hkollmann:
Do you test this with Linux? In my Windows enviroment i do not have any problems with  line endings.

andriy-s:
I'm on Windows. I can reproduce this easily with newly bootstrapped application. If you test with an existing app, then please note that this following works fine for me no matter what line endings are used:

`/**
 * @asset(testapp/*)
 */`

whereas the following fails when using CR/LF:

`/**
 * This is the main application class of "testapp"
 *
 * @asset(testapp/*)
 */`

level420:
@hkollmann it seems that the fix is not working for me, but it may also be that I'm not using the fixed compiler version. Should `1.0.0-beta.20200424-1525` contain https://github.com/qooxdoo/qooxdoo-compiler/commit/023f1d60098d3353c54ab2f9481477163badf0a1 ?

hkollmann:
This is not the fixed version. Something went wrong with travis 

Von meinem iPhone gesendet

hkollmann:
There is a New version online nie

Von meinem iPhone gesendet

level420:
thank you @hkollmann I now have `1.0.0-beta.20200430-1100` which works!

hkollmann:
fixed with https://github.com/qooxdoo/qooxdoo-compiler/pull/665

PR #665, opened 2020-04-28
fix for #662: When a class file has Windows-style line endings (CR/LF…

hkollmann:
…) the @asset annotation does not always work.


johnspackman:
thank you @hkollmann

PR #672, opened 2020-05-02
add some fixes for source maps

hkollmann:
- enable pathMapping in source Maps
- enable relative file names in source Maps
- add saveSourceInMap Option to save source in map file
- fix Schema file

### DIFF
--- a/source/resource/qx/tool/schema/compile-1-0-0.json
+++ b/source/resource/qx/tool/schema/compile-1-0-0.json
@@ -141,7 +141,7 @@
           "minify": {
             "$ref": "#/properties/targets"
           },
-          "saveSourceInMap": {
+          "save-source-in-map": {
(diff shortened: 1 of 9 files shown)


source/resource/qx/tool/schema/compile-1-0-0.json
```diff
@@ -137,6 +137,15 @@
...
           "parts": {
             "$ref": "#/properties/parts"
+          },
+          "minify": {
+            "$ref": "#/properties/targets"
+          },
+          "saveSourceInMap": {
+            "$ref": "#/properties/targets"
+          },
+          "save-unminified": {
```
cboulanger, line 147: shouldn't we be consistent with either CamelCase or hyphen?

johnspackman:
thank you @hkollmann

PR #673, opened 2020-05-04
enhance deploy

hkollmann:
add the possibility declare deploy output dir in compile.json.
With this change it will be possible to define a deploy dir on a target base.
This is necessary if you have more then one build target.


source/class/qx/tool/cli/commands/Compile.js
```diff
@@ -685,7 +685,7 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
         }
 
         // Take the command line for `saveSourceInMap` as most precedent only if provided
-        var saveSourceInMap = targetConfig["saveSourceInMap"] || t.argv["saveSourceInMap"];
+        var saveSourceInMap = targetConfig["save-source-in-map"] || t.argv["save-source-in-map"];
```
johnspackman, line 688: I try to stick to camel case for argv options because it makes it consistent to search - it's just as valid to write `t.argv.saveSourceInMap` because yargs provides both.

Most of the options in compile.json are camelCase, but some are hyphenated; we should stick to a consistent approach, and camelCase is compatible with Javascript code and with Qooxdoo property names.

hkollmann: What's about the hyphen tags in compile .json? Should they changed as well? Breaking compatibility?

johnspackman:
thank you @hkollmann

Issue #633, opened 2020-03-21
Error compiling file `url.replace is not a function`


rad-pat:
Files with a comment such as `    // [Constructor]` within the class definition fail to compile since `1.0.0-beta.20200320-1542` release  
```javascript
qx.Class.define("myapp.ui.myClass", {
    extend: qx.ui.core.Widget,
    // --------------------------------------------------------------------------
    // [Constructor]
    // --------------------------------------------------------------------------
    construct: function()
```

hkollmann:
Found the problem.
You can now use markdown in your comments, 
[Constructor] is a markdown link tag which can not be parsed via Showdown Framework to an valid url.
I will insert some error handling in the markdown cobersion calls.

hkollmann:
Here is the PR: https://github.com/qooxdoo/qooxdoo-compiler/pull/634
Thanks for reporting!

hkollmann:
Done with #634

rad-pat:
Thanks for the quick fix!

johnspackman:
It is reports that this is still happening, by the OP and by @level420 

> @level420 
```
Markdown conversion error: "url.replace is not a function" found in ntrol.ImportList
[Constructor]
```

> @level420 Well I've stripped the file down to

```
qx.Class.define("my.application.control.ImportList",
{
  extend : my.application.control.EditableList,

  members :
  {
    /**
     * Create an editor instance
     *
     * @return {Object}
     */
    _createEditorInstance : function() {
      return null;
    }
  }
});
```
and still get the message. If I remove the doc/comment like this:
```
qx.Class.define("my.application.control.ImportList",
{
  extend : my.application.control.EditableList,

  members :
  {
    _createEditorInstance : function() {
      return null;
    }
  }
});
```
it's gone.

hkollmann:
@derrell : I can not reproduce this with test cases. Could you create a small example project?

rad-pat:
@hkollmann, to reproduce, replace the line comments in the original example to a block comment 
```javascript
qx.Class.define("myapp.ui.myClass", {
    extend: qx.ui.core.Widget,
    /**
    * --------------------------------------------------------------------------
    * [Constructor]
    * --------------------------------------------------------------------------
    */
    construct: function()
```

hkollmann:
[Constructor] is the problem since we use markdown in comments. This is an invalid link. So the message is correct. 
It should be a warning and try to use the text without markdowb support

[...]

PR #676, opened 2020-05-06
#633: Error compiling file `url.replace is not a function

hkollmann:
- markdown error as warning
- add more tests


source/class/qx/tool/compiler/jsdoc/Parser.js
```diff
@@ -99,7 +99,7 @@ qx.Class.define("qx.tool.compiler.jsdoc.Parser", {
           try {
             cmd.body = converter.makeHtml(cmd.body);
           } catch (e) {
-            qx.tool.compiler.Console.error(`Markdown conversion error: "${e.message}" found in \n${cmd.body.trim()}`);
+            qx.tool.compiler.Console.warn(`Markdown conversion: "${e.message}" found in \n${cmd.body.trim()}`);
```
johnspackman, line 102: Isn't the exception being raised a bug in the "showdown" module?  EG what is wrong with having a comment like 

```
    /**
     * --------------------------------------------------------------------------
     * [Constructor]
     * --------------------------------------------------------------------------
     */
    construct: function()
```

Even if we decided to enforce strict rules on comment formatting, it should not say something inexplicable just `url.replace is not a function`.

cboulanger: Maybe the simplest thing would be making the warning more explicative: 
```
Markdown conversion problem: Error "${e.message}" was thrown parsing "${cmd.body.trim()}". Please review your doc comments for Markdown compatibility.
```

derrell: I don't think there should be *anything* in a comment that causes it to barf. If it's understandable as markdown, then treat it as such. If not, don't tell the user how to write a comment. (A warning that can be turned off via command-line option or `@ignore` or whatever is ok; just not an error.)

hkollmann:
Implemented both suggestions from @cboulanger  and @derrell

level420:
BTW: this `url.replace is not a function` seems to be a bug in node module showdown: https://github.com/showdownjs/showdown/issues/548
Created in 2018, labeled as a bug, but still unresolved :-(
So we are fighting agains a bug in another module, which is bad.

level420:
and what I've understood from that bug is, that every markdown named like a javascript reserved word could cause the exception, like e.g. constructor

level420:
@johnspackman as I've understood it will only break for links, but its bad anyway

hkollmann:
Do you know an alternative to showdown?

johnspackman:
Not personally but there are several out there.

Github use https://github.com/github/cmark-gfm, but that's a C library so there is this nodejs wrapper: https://github.com/killagu/node-cmark

For a pure JS approach there is https://github.com/markdown-it/markdown-it "Markdown parser done right. Fast and easy to extend."

Or https://marked.js.org/

cboulanger:
@hkollmann Can you document why did you closed this? Was the bug resolved with another PR?

hkollmann:
was an accident

cboulanger:
@johnspackman You are right, but can we accept this workaround now and deal with switching the markdown parser later, since this would push the release date again? The current system is working except this bug. I'll create a separate issue for v7 so we don't forget. Ok?

johnspackman:
Thank you @hkollmann

PR #677, opened 2020-05-07
fixes a regression in deploy

johnspackman:


cboulanger:
Thank you @johnspackman for fixing this at light speed!

PR #679, opened 2020-05-08
Fix #663 (qx.debug should be false for the "build" target)

### LINKED ISSUE #663: The qx.debug setting in the environment should be false when compiling the build target
The `qx.debug` setting in the environment seems to be always `true` by default, no matter what target is being compiled. While explicitly overriding it in the `compile.json` for a specific target works for me, it should instead be implicitly false for the `build` target.

andriy-s:


cboulanger:
How does this work, though? Your are just "getting" the `defaultEnvironment` property - nothing is done with its value. Or do I miss something?

andriy-s:
> Your are just "getting" the `defaultEnvironment` property - nothing is done with its value.
The diff viewer does not show enough lines to get the full context, try expanding it up. The value that this.getTarget().getDefaultEnvironment() returns is then passed as an additional parameter to qx.tool.utils.Values.merge() to composed the compileEnv object. And the order of calls is such that the this.getTarget().getEnvironment(), which holds the environment defined in the compile.json file, can still override the target type default.
Does this answer your question, or did I understood it wrong?

cboulanger:
Sorry, of course! Didn’t look closely enough.

cboulanger:
Thank you @andriy-s !

PR #680, opened 2020-05-10
Fixed error message that mentions deployDir parameter whereas it shou…

andriy-s:
…ld be deployPath instead.

While here, enhance the message so it is clear in what place in the compile.json this property should be placed.


johnspackman:
thank you @andriy-s

PR #681, opened 2020-05-12
Fix the yargs docs for qx serve and qx test

cboulanger:
This should also fix the regression that `qx test` isn't using `compile-test.json` as default config file (if present).


johnspackman:
thank you @cboulanger

PR #683, opened 2020-05-12
Fix loader on IE11

andriy-s:


cboulanger:
Merged, thank you @andriy-s !

PR #685, opened 2020-05-13
Sort package cache data by release date

cboulanger:
`published_at` will be used for sorting the cache so that the diffs become easier to read.

### DIFF
--- a/source/class/qx/tool/cli/commands/package/Install.js
+++ b/source/class/qx/tool/cli/commands/package/Install.js
@@ -282,7 +282,7 @@ qx.Class.define("qx.tool.cli.commands.package.Install", {
       if (!release_data) {
         throw new qx.tool.utils.Utils.UserError(`'${repo_name}' has no release '${tag_name}'.`);
       }
-      // TODO: the path in the cache data should be the path to the library containing Manifest.json, not to the Manifest.json itself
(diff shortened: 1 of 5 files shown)


cboulanger:
Do not merge yet, the sorting bit is still missing ...

cboulanger:
@hkollmann @johnspackman PR is now ready for review & merge

source/class/qx/tool/cli/commands/package/Install.js
```diff
@@ -282,7 +282,7 @@ qx.Class.define("qx.tool.cli.commands.package.Install", {
       if (!release_data) {
         throw new qx.tool.utils.Utils.UserError(`'${repo_name}' has no release '${tag_name}'.`);
       }
-      // TODO: the path in the cache data should be the path to the library containing Manifest.json, not to the Manifest.json itself
+      // TO DO: the path in the cache data should be the path to the library containing Manifest.json, not to the Manifest.json itself
```
hkollmann, line 285: Should stay TODO so that it can be found by tools

cboulanger: I wanted to remove the linter warnings since the todo is not really important. The whole stuff needs to be rewritten one day, and then the todos will be meaningless.

Issue #476, opened 2019-06-28
scramble private method names (__methodName)


oetiker:
I just ran into a case in our inhouse code where people were happily calling `__privateMethods` from the parent in a child class ... it would be really good if the compiler could start scrambling these names asap ... or we will have a major issue at our hands re-introducing this later on ...

hkollmann:
You can activate a lint rule to check this: @qooxdoo/qx/no-illegal-private-usage 
It's not on by Default because it's very very slow.
Source is here https://github.com/qooxdoo/eslint-qx-plugins.

hkollmann:
@oetiker: I refactored the no-illegal-private-usage rule a little bit so that is usable now.
Could you try if it is working for you? 
Here is the PR to update the compiler.

oetiker:
the problem is that if we do this with the linter only, people will still use __ accessors, maybe just eslint exceptions to their code. @johnspackman is there anything hindering the compiler from scrambling the names as the old compiler used to.

hkollmann:
IMHO that is some sort of organization to forbid this, is'nt it? Problem with scrambling is - also in the old compiler - that you get the errors only with builds - not with debug code.
I often had this problem by myself. 
Therefore i like the no-illegal-private-usage rule. Just fixed my own code and the compiler code see https://github.com/qooxdoo/qooxdoo-compiler/pull/481

Most errors in compiler and my was the missing of the declaration of the private members.

oetiker:
private members are a feature that had been present in qooxdoo oo for a long time ... and yes the old compiler did not do it in source mode, but now with the new compiler, we can do this ... since the code does always get compiled. Scrambling I guess will also take care of the problem that a private property in the child should NOT affect a private property of the same name in the parent.

johnspackman:
@oetiker 's right, it's something that the generator used to do and which I've always kind of planned but never got round to.

Part of the reason for my reticence was the effort it takes to correctly statically analyse code, but thinking about now there might be an easier option: for any variable prefixed with `__`, just rename it to `__` plus a random ID which is pseudo-unique (enough to avoid collisions most of the time).  This would be little more than a search & replace without analysing the context of the reference (which might be what the generator used to do).

Sometimes strings have to be mapped too - eg `this.disposeObjects("__myPrivateVar")` is a common example.

I agree it should work in source mode too - although in source mode that the unique ID should be a suffix rather than a replacement for debugging.

PR #693, opened 2020-05-15
Privates mangling

johnspackman:
This adds mangling for private symbols, which work in source targets as well as build targets.

There is a PR in the framework https://github.com/qooxdoo/qooxdoo/pull/9973 which goes along side this one - now that names are mangled, part of this PR is to make sure that the display name of the function is preserved as the original, and until that PR is merged into the framework you will see some ugly names in the call stack


johnspackman:
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/476

source/class/qx/tool/compiler/Analyser.js
```diff
@@ -123,6 +123,12 @@ qx.Class.define("qx.tool.compiler.Analyser", {
       nullable: false,
       check: "Array"
     },
+    
+    /** Whether and how to mangle private identifiers */
+    manglePrivates: {
+      init: "readble",
```
hkollmann, line 129: Here is a typo...

johnspackman: oh yes - and also a bug that the framework allows it!

cboulanger:
Thank you @johnspackman !

PR #703, opened 2020-05-23
Fix #702: Translation strings are only taken from main app

### LINKED ISSUE #702: Translation strings are only taken from main app
If you use a library, e.g. qxl.Dialogs, which have its own translations, they will not be used. Only translations from the main app are used.

From Target.js:
```
_writeAllTranslations: async function(appMeta) {
      var t = this;
      var analyser = appMeta.getAnalyser();
      let bootPackage = appMeta.getPackages()[0];

      var promises = t.getLocales().map(async localeId => {
        let translation = await analyser.getTranslation(appMeta.getAppLibrary(), localeId);
        let pkg = this.isI18nAsParts() ? appMeta.getLocalePackage(localeId) : bootPackage;
        var entries = translation.getEntries();
        for (var msgid in entries) {
          pkg.addTranslationEntry(localeId, entries[msgid]);
        }
      });
      await qx.Promise.all(promises);
    },
```
appMeta.getAppLibrary() get only main app library. 

Here it's simple.  _writeRequiredTranslations has the same problem.

hkollmann:


johnspackman:
thank you @hkollmann

PR #705, opened 2020-05-26
More scalable support for multiple event listeners 

cboulanger:
The tests run by `qx test` are are implemented as event listeners to the `runTests`  event. This means that event listeners can be registered form various places, such as the main app and installed packages. This async solution is very flexible and perfectly fits the architecture of the compiler, The drawback is that using events makes it difficult to get the result of the tests, since events are one-way constructs. 

Currently, the `runTests` event is fired with a simple javascript object, and tests are expected to write to the `errorCode` property of that object. This is difficult to document & a bit un-qooxdooish. Also, tests have to take care of handling previous `errorCode`s since simply overwriting the property will lose any existing test results (if I read the code correctly).

This PR introduces a slightly more convoluted, but clearly documentable and qooxdoo-style solution. Tests call `qx.tool.cli.commands.Test.prototype.registerTest()` with an instance of `qx.tool.cli.api.Test` and on completion (or failure) of the test, set the `exitCode` qx property of that instance. The instance can also have additional information such as the name of the Test. 

The previous way is still supported, since instead of setting the `errorCode` property of a passed javascript object, the test case will now simply set the corresponding property of the command instance. 

This PR also makes it possible to skip starting a server, which is not necessary for server-side tests.

### DIFF
--- a/compile.js
+++ b/compile.js
@@ -1,0 +1,59 @@
+qx.Class.define("qx.compiler.LibraryApi", {
+  extend: qx.tool.cli.api.LibraryApi,
+
+  members: {
+    async load() {
+      let command = this.getCompilerApi().getCommand();
+      if (command instanceof qx.tool.cli.commands.Test) {
+        command.addListener("runTests", this.__test1, this);
+        command.addListener("runTests", this.__test2, this);
+        command.addListener("runTests", this.__test3, this);
+      }
+    },
+
+    /**
+     * @param {qx.event.type.Data} evt Event which has the current command instance
+     * of type {@link qx.tool.cli.commands.Test} as data.
+     * @private
+     */
+    async __test1(evt) {
+      const cmd = evt.getData();
+      const test = new qx.tool.cli.api.Test("Test 1");
+      cmd.registerTest(test);
+      // do some async testing and set exit code explicitly
+      await new qx.Promise(resolve =>  qx.event.Timer.once(resolve, null, 1000));
+      test.setExitCode(0);
+    },
+
+    /**
+     * @param {qx.event.type.Data} evt
+     * @private
+     */
+    async __test2(evt) {
+      const cmd = evt.getData();
+      const test = new qx.tool.cli.api.Test("Test 2");
+      cmd.registerTest(test);
+      // do some async testing
+      await new qx.Promise(resolve =>  qx.event.Timer.once(resolve, null, 1000));
+      qx.tool.compiler.Console.info(`The next test fails on purpose.`);
+      test.setExitCode(255);
+    },
+
+    /**
+     * @param {qx.event.type.Data} evt
+     * @private
+     */
+    async __test3(evt) {
+      const cmd = evt.getData();
+      const test = new qx.tool.cli.api.Test("Test 3");
+      cmd.registerTest(test);
+      // do some async testing
+      await new qx.Promise(resolve =>  qx.event.Timer.once(resolve, null, 1000));
+      test.setExitCode(0);
+    },
+  }
+});
+
+module.exports = {
+  LibraryApi: qx.compiler.LibraryApi
+};
--- a/source/class/qx/tool/cli/Cli.js
+++ b/source/class/qx/tool/cli/Cli.js
@@ -35,35 +35,35 @@ qx.Class.define("qx.tool.cli.Cli", {
     }
     qx.tool.cli.Cli.__instance = this;
   },
-  
+
   members: {
     /** @type {yargs} the current yargs instance */
     yargs: null,
-    
+
     /** @type {Object} the current argv */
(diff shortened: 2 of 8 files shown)


cboulanger:
BTW, we could move all javascript tests currently run using shell scripts into the scope of `qx test` and reduce the `npm test` to CLI tests (which then implicitly call `qx test`)

cboulanger:
And I forgot - this PR also removes installation of missing libraries in the case of `qx clean`, which made no sense whatsoever.

cboulanger:
Sorry for the whitespace changes, I need to switch that off in my IDE. In case you don't know that already, can hide them like so: https://github.blog/2018-05-01-ignore-white-space-in-code-review/

compile.js
```diff
@@ -0,0 +1,59 @@
+qx.Class.define("qx.compiler.LibraryApi", {
```
hkollmann, line 1: Is'nt this test code? If yes it should be somwhere in the test directory.

source/class/qx/tool/cli/Cli.js
```diff
@@ -360,8 +360,10 @@ qx.Class.define("qx.tool.cli.Cli", {
           config.packages[library.uri] = library.path;
         });
       }
-
-      if (config.libraries) {
+      // check if we need to load libraries, needs more robust test
+      let needLibraries = qx.lang.Type.isArray(this.argv._) && this.argv._[0] !== "clean";
```
hkollmann, line 364: upgrade do not need loading libraries either

source/class/qx/tool/cli/api/Test.js
```diff
@@ -0,0 +1,26 @@
...
+      event: "changeName"
+    },
+    description: {
+      check: "String",
+      event: "changeDescription"
+    },
+    exitCode: {
+      check: "Number",
+      event: "changeExitCode",
+      nullable: true,
```
hkollmann, line 22: Should be 0 as default and not nullable. It's the process exit code should be a  valid value.

source/class/qx/tool/cli/commands/Test.js
```diff
@@ -20,7 +20,20 @@ const path = require("path");
 const process = require("process");
 
 /**
- * Compiles the project and serves it up as a web page
+ * Compiles the project, serves it up as a web page (default, can be turned off),
```
hkollmann, line 23: With this approach registered test do not need an explicit event. What's about listening for the runaTest event here and execute a runTest method for all registered tests?

compile.js
cboulanger: It is, but you don't run any `qx` in the repo root of qooxdoo-compiler repo unless you are doing some testing or development - so I don't think it hurts. Tests will only be run when `qx test` is executed.

hkollmann: i run qx lint in the repo root!

cboulanger: Ideally, this would run "real" tests, i.e. the tests in the test dir. These are just stubs to test the functioning of the "internal" test runner itself. Should be replaced with something real eventually.

hkollmann: ok, agreet.

source/class/qx/tool/cli/api/Test.js
cboulanger: If it is 0 then you cannot listen for the changeExitCode event in case of a successful test. It needs to be in some "undefined" state to signal the test hasn't run yet. If it is not set, the test hasn't run.

source/class/qx/tool/cli/commands/Test.js
cboulanger: Can you explain (maybe with code)? I don't fully understand...

source/class/qx/tool/cli/Cli.js
cboulanger: True!

source/class/qx/tool/cli/commands/Test.js
hkollmann: Will add a change to pr later ok?

cboulanger: Sure!

cboulanger: Or accept-as-is and create new PR`? Might be easier to follow.

cboulanger:
@hkollmann @johnspackman Would you accept the PR so that John's fix is in the code - further modifications can be done in separate PRs? Or alternatively, John moves the fix into its own PR which we can fast-track?

johnspackman:
thank you @cboulanger

source/class/qx/tool/cli/Cli.js
cboulanger: Can you add that in your PR?

PR #706, opened 2020-05-27
fixes a bug in mangling

johnspackman:


cboulanger:
Thank you @johnspackman !

PR #709, opened 2020-05-30
Support self-compiling compiler & enhance test api

hkollmann:
This PR should simplify generating tests to simple 

```
command.addTest(new qx.tool.cli.api.Test("Test 0", () => {
   tesccode 
}));;
```
Example found in test/testTestApp

Some work is left:
   - waiting for https://github.com/qooxdoo/qooxdoo/pull/9986
   - move some of the compiler tests to the new api.
 
This PR is to discuss the api in an early stage. So please do this!


source/class/qx/tool/cli/commands/Test.js
```diff
@@ -47,36 +45,10 @@ qx.Class.define("qx.tool.cli.commands.Test", {
     CONFIG_FILENAME : "compile-test.json",
 
     YARGS_BUILDER: {
-      verbose: {
```
cboulanger, line 50: Why did you remove all these parameters? Are they duplicates?

cboulanger:
I like it!

source/class/qx/tool/cli/commands/Test.js
hkollmann: Because they are special to the testtapper package. So testapper should add them. With the new api this should be possible as john said

source/class/qx/tool/cli/commands/Test.js
```diff
@@ -22,18 +22,16 @@ const process = require("process");
...
+ * 
+ * The runTests event is called with a {@link qx.even.type.Data}
  * containing the command instance.
  *
  * A test can write to the (native) `errorCode` property of this instance,
  * which is used to determine the exit code of the `qx test` command.
  * This means, however, that the last run test overwrites the
  * `errorCode` of any previous test, and has to take care of this.
  *
- * A more scalable solution is to  call {@link
```
cboulanger, line 33: Shouldn't it be the other way round - remove the docs on `errorCode` which is only there for BC purposes, and tell people to use the proper way with the `qx.tool.cli.api.Test`, since this is the cleaner solution which doesn't require boilerplate to check existing exitCodes? Also, this supports the `--fail-fast` option which `errrorCode` doesn't (without additional checks).

source/class/qx/tool/compiler/Version.js
```diff
@@ -28,7 +28,11 @@ qx.Class.define("qx.tool.compiler.Version", {
   },
 
   defer: function(statics) {
-    var pkg = require("../../../../../package.json");
-    statics.VERSION = pkg.version;
+    try {
+      var pkg = require("../../../../../package.json");
+      statics.VERSION = pkg.version;
+    } catch (e) {
+      statics.VERSION = qx.core.Environment.get("qx.compilerVersion");
```
cboulanger, line 35: maybe better call it `qx.compiler.version`? to avoid camelCase in config key names?

hkollmann:
It is a typo thanks
Not ready yet - but a step forward 
I underestimated the work 

Von meinem iPhone gesendet

source/class/qx/tool/cli/commands/Compile.js
```diff
@@ -224,8 +228,23 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
...
+    /**
+     * Fired when minification begins, data is a map containing:
+     *  application {qx.tool.compiler.app.Application} the app being minified
+     *  part: {String} the part being minified
+     *  filename: {String} the part filename
+     */
+    "minifyingApplication": "qx.event.type.Data",
 
+    /**
+     * Fired when minification begins, data is a map containing:
```
cboulanger, line 242: s/begins/is done/  maybe?

hkollmann:
That's it. qx compiles and test itselfs.
Inclusive unit testing. Last thing to do is to convert the test in test/_todo_move_to_unit_test to unit tests.

Blackbox box compiler test are done via tape and could be found in test/compiler.
All bats scripts are converted - bats is removed.

Example for an converted unit test is found in qx.test.tool.utils.Utils.
Unit tests are run through the new qxl.testnode package.

cboulanger:
@hkollmann Thank you very much, Henner, for this massive piece of work. It was probably much more than expected. I am thrilled we've achieved this before release!

source/class/qx/test/compiler/jsdoc/Parser.js
```diff
@@ -0,0 +1,68 @@
...
+   License:
+     MIT: https://opensource.org/licenses/MIT
+     See the LICENSE file in the project's top-level directory for details.
+
+   Authors:
+   * Henner Kollmann 
+
+************************************************************************ */
+
+qx.Class.define("qx.test.tool.utils.Utils", {
```
cboulanger, line 16: The filename (`qx/test/compiler/jsdoc/Parser.js`) differs from the class name (`qx.test.tool.utils.Utils`)

source/class/qx/tool/cli/commands/Test.js
cboulanger: I would still argue that we should encourage people to use the `exitCode` qooxdo property of the Test instance instead of the `errorCode` native property ...

source/class/qx/tool/cli/commands/Test.js
```diff
@@ -156,29 +151,35 @@ qx.Class.define("qx.tool.cli.commands.Test", {
...
       }
       this.addListener("making", () => {
-        if (!this.hasListener("runTests")) {
+        if (!this.hasListener("runTests") && (this.__tests.length === 0)) {
           qx.tool.compiler.Console.error(
-            `No test runner registered!
+            `No tests are registered!
                Please register a testrunner, e.g. testtapper with:
                qx package install @qooxdoo/qxl.testtapper
+               See documentation https://qooxdoo.org/docs/#/development/testing/
```
cboulanger, line 165: See documentation **at** https:/

source/class/qx/tool/cli/commands/Test.js
```diff
@@ -156,29 +151,35 @@ qx.Class.define("qx.tool.cli.commands.Test", {
...
       });
-      this.__tests = new qx.data.Array();
-      this.addListener("afterStart", () => {
-        let res = this.fireDataEvent("runTests", this);
-        res.then(() => {
-          process.exit(this.errorCode);
-        });
+
+      this.addListener("afterStart", async () => {
+        qx.tool.compiler.Console.info(`run unit tests`);
```
cboulanger, line 173: I would output "Running unit tests"

johnspackman:
@hkollmann a great addition!

this is a huge PR though, with lots of changes - I've merged your branch into my working master so that I can try it out, but is the `lib` directory missing from the repo?  I get this:

```
$ pwd
/Users/john/dev/Qooxdoo/QxCompiler
$ ./bin/qx
internal/modules/cjs/loader.js:983
  throw err;
  ^

Error: Cannot find module '/Users/john/dev/Qooxdoo/QxCompiler/lib/compiler'
Require stack:
- /Users/john/dev/Qooxdoo/QxCompiler/bin/qx
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:980:15)
    at Function.Module._load (internal/modules/cjs/loader.js:862:27)
    at Module.require (internal/modules/cjs/loader.js:1042:19)
    at require (internal/modules/cjs/helpers.js:77:18)
    at Object.<anonymous> (/Users/john/dev/Qooxdoo/QxCompiler/bin/qx:4:1)
    at Module._compile (internal/modules/cjs/loader.js:1156:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1176:10)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/john/dev/Qooxdoo/QxCompiler/bin/qx' ]
}

```

hkollmann:
That is as expected. Compiler needs to be compiled first. Just run `qx deploy --target=build` .

I have to update the readme.

johnspackman:
> That is as expected. Compiler needs to be compiled first. Just run `qx deploy --target=build` .

but you can't because `qx` does not exist...

IMHO during development, there needs to be a dummy project which depends on the current `@qooxdoo/compiler`; then add a command (eg) `qxqx` in the root needs to be a link to the `./dummyproject/node_modules/@qooxdoo/compiler/bin/qx`.  This allows `qx` to be whatever is the currently compiled, local copy of the compiler and `qxqx` to be the published version.

hkollmann:
Or you just install the current compiler global.

johnspackman:
my compiler is global, I only use one instance, and that's the one which is being coded :)

johnspackman:
the difference is that with the `qxqx` approach, *everything* (ie not just the compiler) is being compiled by the bleeding edge, including all client apps in test or production.  This adds lots of real world testing, for free - it's the "eating your own dog food" strategy

The `qxqx` would only be used to compile that bleeding edge compiler.

If you have to install a global compiler, everything is compiled by the global instance and it misses out on all of the real world testing otherwise available, and you have to remember to update it etc.  It's always out of sync with the new features you're coding into the compiler.

cboulanger:
@johnspackman This is already possible, isn’t it? Or does Henner need to change anything? Like requiring the NPM Compiler as a devDependency?

cboulanger:
Of course, the devDependency would be removed before publishing

johnspackman:
> Like requiring the NPM Compiler as a devDependency?

I doubt you can do that (although I've not tried) because it would mean a self-dependency and I'd guess that npm's behaviour would be at best undefined

> Of course, the devDependency would exist only in Master an be removed before publishing

That would mean constantly editing the package.json before commit and after checkout.  

I like the idea of a dummy project - it's entirely predictable and understandable, and is easily created/rebuilt automatically during `git checkout`

hkollmann:
devDependencies is working. I am just testing what's happening with npm pack.
The devDependencies has the advantage that you only need one npm install after checkout.
With the dummy project you need 2 calls: One in the dummy project and one in the compiler

source/class/qx/tool/cli/commands/Test.js
hkollmann: Oh, errorCode is gone. Testtapper has to be fixed after release

source/class/qx/test/compiler/jsdoc/Parser.js
hkollmann: You are right. Was a start to concert the missing tests

hkollmann:
bootstrap compile can now be done with 
```
npm install
npx qx deploy --target=build
```

cboulanger:
If the `devDependency` was a problem after all, I think it would be very easy for a deploy script to remove it before publish, something like 
```
echo "$(jq '.devDependencies = {}' package.json)" > package.json
```

cboulanger:
@hkollmann Why do we have to use `--target=build`? Isn't this the default when `qx deploy` is used?

hkollmann:
no it is'nt

johnspackman:
great!  I've merged it with my master and will be using it from today

cboulanger:
This is totally awesome. Great job, @hkollmann !

cboulanger:
I'll be able to test it this afternoon. Probably it would be good to test it on all qxl.* apps, just to be sure, before merging? (Since this is such a massive PR).

hkollmann:
I thought therefore we have the https://github.com/qooxdoo/deployment?

cboulanger:
When compiling the compiler, I get los of warnings about unrecognized images and a few unresolved symbol warnings - I assume you get them, too. Compilation of my app worked flawlessly.

hkollmann:
> When compiling the compiler, I get los of warnings about unrecognized images and a
Did'nt see this yet. Do you have an log?

> few unresolved symbol warnings - I assume you get them, too. Compilation of my app worked flawlessly.
Needs to be fixed in the framework

cboulanger:
> I thought therefore we have the https://github.com/qooxdoo/deployment?

Yes, but we can easily automate this using
```
"scripts": {
    "prepublish": "echo \"$(jq '.devDependencies = {}' package.json)\" > package.json"
  },
```
This should work, shouldn't it?

cboulanger:
> Did'nt see this yet. Do you have an log?

https://pastebin.com/iNTAhVsR

hkollmann:
Seems to be an mac problem: https://stackoverflow.com/questions/45004352/error-enfile-file-table-overflow-scandir-while-run-reaction-on-mac

johnspackman:
ah - looks like the resource manager needs to pool, instead of opening unlimited number of images.  This should be a separate bug (async file I/O is great, but there can't be much benefit from going async on 1,000's of files).

cboulanger:
@hkollmann Upping the number of open files on the Mac (as suggested in the Stackoverflow question) did the trick, I am down to 
```
qx.bom.client.Pdfjs: [399,18] warning: Unresolved use of symbol $.browsera.ApplicationMeta
qx.dev.unit.Sinon: [4004,19] warning: Unresolved use of symbol DOMParser
qx.dev.unit.Sinon: [4008,25] warning: Unresolved use of symbol ActiveXObject
qx.ui.root.Application: [132,10] warning: Unresolved use of symbol alert
qx.bom.History: [331,39] warning: Unresolved use of symbol history.back
qx.bom.History: [340,39] warning: Unresolved use of symbol history.forward
```
But this shouldn't be necessary, so using pooling as suggested by @johnspackman would be good. Should I open an issue or are you planning on doing it, @johnspackman ?

hkollmann:
And this has to be fixed in the framework

cboulanger:
Updated the PR title because it has moved well beyond its initial scope.

What about adding these self-executing scripts:
```
"scripts": {
   "postinstall": "npx qx deploy --target=build",
   "prepublish": "echo \"$(jq '.devDependencies = {}' package.json)\" > package.json"
  },
``` 
This makes sure the package is *never* published to NPM with a recursive dependency. 
The only place the prepublish hook is executed is in the workflow script in the deployment
repository, since apart from very exceptional circumstances, the package should never be 
published manually. 

What we still need is a workflow step in this repo that triggers the scripts in the deployment
repository.

hkollmann:
Are you shure jq is installed?

hkollmann:
what's about https://github.com/FGRibreau/jq.node as devDependency? Would work in windows too

cboulanger:
> what's about https://github.com/FGRibreau/jq.node as devDependency? Would work in windows too

sure, this should work with `npx jq` without having to install it...

.vscode/launch.json
```diff
@@ -0,0 +1,95 @@
+{
```
cboulanger, line 1: should probably not be committed? .vscode -> .gitignore ?

source/class/qx/tool/cli/commands/Deploy.js
```diff
@@ -199,10 +219,10 @@ qx.Class.define("qx.tool.cli.commands.Deploy", {
 
   defer: function(statics) {
     qx.tool.compiler.Console.addMessageIds({
-      "qx.tool.cli.deploy.deployDirNotSpecified": "No deploy dir configured! Use --out parameter or deployPath target property in compile.json."
+      "qx.tool.cli.deploy.deployDirNotSpecified": "No deploy dir fot target <%1> configured! Use --out parameter or deployPath target property in compile.json."
```
cboulanger, line 222: s/fot/for/

hkollmann:
Do you have a Problem with vscode? It is in the gitignore. Just a helper config

Von meinem iPhone gesendet

cboulanger:
@hkollmann I thought it was commited by mistake, just ignore my comment

source/class/qx/tool/cli/commands/Deploy.js
```diff
@@ -72,10 +72,10 @@ qx.Class.define("qx.tool.cli.commands.Deploy", {
      * fired after deploying. With this event 
      * application can do aditional copying 
      *
```
cboulanger, line 74: It‘s a bit unclear what this documents- I assume the properties of an object in the event data ? Should be spelled out more clearly.

hkollmann: Yes it's the object in the event data. Any idea how to make it clearer?

hkollmann:
ApiViewer know can document qooxdoo server apps!
See new enviroment key buildApiForApp.

source/class/qx/tool/cli/commands/Deploy.js
cboulanger: Just add that „the event data is an object with the following properties:“

johnspackman:
thank you @hkollmann 

PS - merging this despite travis because it's been 3 days and travis has not responded (sigh)

PR #712, opened 2020-06-04
Deploy an API Viewer for qx.tool.* on GH pages

cboulanger:


cboulanger:
@hkollmann You're right, that was a mistake! Thanks for catching it.

PR #719, opened 2020-06-08
npm link overwrite framework with current  npm version 

hkollmann:
- so test in framework will not use the just copied version


johnspackman:
Thank you @hkollmann

PR #720, opened 2020-06-09
Change let to var in template to ensure backward compatibility

rommni:
Hi,
As said on gitter, like this let is not impact by user babel setting it prevents users to do old browser compatibility (broswer without let support).


cboulanger:
Thank you @rommni !

PR #721, opened 2020-06-11
adds `--output-path-prefix` to `qx compile` (needed for deployment)

johnspackman:


source/class/qx/tool/cli/commands/Compile.js
```diff
@@ -642,6 +646,8 @@ qx.Class.define("qx.tool.cli.commands.Compile", {
         }
 
         var outputPath = targetConfig.outputPath;
+        if (this.argv.outputPathPrefix)
+          outputPath = path.join(this.argv.outputPathPrefix, outputPath);
         if (!outputPath) {
```
derrell, line 651: This looks like it is possible to specify `--output-path=/x/y/` and `--out=` (empty value) or `--output-path=/` and `--out=x/y/` to yield the same result. What is the difference between `--output-path` and `--out`? The latter was already a directory...???

johnspackman: where are you seeing the `--out`?  

The idea behind `--output-path-prefix` is that you can add a prefix to the output directory which is specified in the `compile.json`'s target definition; the only use case for doing this is as part of bootstrapping deployment, because of recursive dependencies between compiler and framework (ie this allows the compiler to compile a version of the framework, when the compiler is dependent on the framework code itself, and without overwriting the files that the compiler is executing)

derrell: Oh, it's different use than I thought. I saw "needed for deployment" and thought it applied to the `deploy` command, which has `--out`.

PR #727, opened 2020-06-29
supports running `qx` from source (ie not requiring a `qx deploy` stage)

johnspackman:
the files in `bin-src/` are just a simple way to use the compiled version of `qx`


bin-src/qx
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+const path=require("path");
+const updateNotifier = require("update-notifier");
+const pkg = require(path.join(__dirname, "..", "package.json"));
+updateNotifier({pkg}).notify({defer:false});
```
hkollmann, line 5: IMHO you do not need the update notification here. It's for running the developmnent verion only, is'nt it?

johnspackman: yep

cboulanger:
Travis seems to be permanently broken, merging

PR #735, opened 2020-07-23
Small fixes

hkollmann:
- move require to AbstractApi to use it from CompilerApi context
- use CompilerApi in compile.js. Tests should only run in compile context not in library

### DIFF
--- a/source/class/qx/tool/cli/api/CompilerApi.js
+++ b/source/class/qx/tool/cli/api/CompilerApi.js
@@ -59,11 +59,11 @@ qx.Class.define("qx.tool.cli.api.CompilerApi", {
     __libraryApis: null,
     
     /**
-     * called before testing starts
-     * 
-     * @param cmd {qx.tool.cli.commands.Test} the current used command instance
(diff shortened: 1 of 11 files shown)


source/class/qx/tool/cli/api/CompilerApi.js
```diff
@@ -68,7 +93,7 @@ qx.Class.define("qx.tool.cli.api.CompilerApi", {
         config = await qx.tool.utils.Json.loadJsonAsync(compileJsonPath);
       }
       this.setConfiguration(config);
-      return config;
```
oetiker, line 71: this breaks existing compile.js files as like mine which relies on load returning the current config ... now it returns the result of the abstract load which is nothing  ... is there a purpose to this breakage ?

hkollmann: It returns the promise of async load.
load could be async now - therefore can return a Promise.

Could you change to 
```
await load();
config = this.getConfiguration();
```

PR #736, opened 2020-07-26
Add node paths

hkollmann:
- fixes an issue if node scripts in script dir contains a require(Charts.js) where charts js ´can be found in script dir and is loaded before.

- fixes loading of modules loaded in compile.js

- add some error handling for tests


source/resource/qx/tool/cli/templates/loader/loader-node.tmpl.js
```diff
@@ -1,6 +1,22 @@
 (function(){ 
 
   var path = require("path");
+  var module = require("module").Module;
+
+  var appModulePaths = [];
+  var old_nodeModulePaths = module._nodeModulePaths;
```
johnspackman, line 7: is `_nodeModulePaths` an undocumented node feature?

source/resource/qx/tool/cli/templates/loader/loader-node.tmpl.js
```diff
@@ -1,6 +1,22 @@
...
+  module._nodeModulePaths = function(from) {
+    var paths = old_nodeModulePaths.call(this, from);
+    paths = paths.concat(appModulePaths);
+    return paths;
+  };
+
+  function addNodePath(dir) {
```
johnspackman, line 14: isn't this going to be quite expensive during startup?  eg when `allScripts` is quite large?

source/resource/qx/tool/cli/templates/loader/loader-node.tmpl.js
hkollmann: found it here: https://www.npmjs.com/package/app-module-path
It's an exported function

source/resource/qx/tool/cli/templates/loader/loader-node.tmpl.js
hkollmann: allScripts are only the scripts from urisBefore, so the list is quite small

 var allScripts = l.decodeUris(l.urisBefore, "resourceUri");
 t.loadGobalPaths(allScripts);
 t.loadScriptList(allScripts);

hkollmann:
The intention is to add the dirs from the externalResources, script section.

  "externalResources": {
    "script": [
      "chart.js/moment.js",
      "chart.js/Chart.js",
      "chart.js/chartjs-plugin-datalabels.js"
    ],
  },

In this example chartjs-plugin-datalabels.js contains a require("Chart.js") which is not part of the global node installation but found in the external resource dir chart.js.

Without adding charts.js to the search path you get an error during loading of chartjs-plugin-datalabels.js.

johnspackman:
> The intention is to add the dirs from the externalResources, script section.
> 
> "externalResources": {
> "script": [
> "chart.js/moment.js",
> "chart.js/Chart.js",
> "chart.js/chartjs-plugin-datalabels.js"
> ],
> },
> 
> In this example chartjs-plugin-datalabels.js contains a require("Chart.js") which is not part of the global node installation but found in the external resource dir chart.js.
> 
> Without adding charts.js to the search path you get an error during loading of chartjs-plugin-datalabels.js.

Sorry, I still dont underdstand why this is necessary - the file `chart.js/chartjs-plugin-datalabels.js` should be doing `require("./Chart.js")` and that should have fixed all search path issues.  It's not our issue to solve the load order for included modules, but it sounds like that's what we would be doing here?

hkollmann:
chartjs-plugin-datalabels.js is an external library which need to be patched in this case. I will try this.

hkollmann:
You are right - patching the load order is the better way in this case.

Issue #737, opened 2020-07-28
Compiler version change from 1.0.0-beta.... to 2.0.0 after qx package migrate 


rommni:
I tested the last version of compiler (last commit: https://github.com/qooxdoo/qooxdoo-compiler/commit/d144c1498ece2e47cc8708aa1280a9e5448978b5) by doing git clone and npm pack. 

After running npm install with new tar.gz generated i run qx compile and it asked me : 
```
*** Framework and/or compiler dependencies in Manifest need to be updated.
*** Please run 'qx package migrate' to apply the changes. If you don't want this, downgrade to a previous version of the compiler.
```

So I run `npx qx package migrate` and it changed in my Manifest.json the following lines:
```
  -  "@qooxdoo/framework": "^6.0.0-beta-20190617-1533",
  -  "@qooxdoo/compiler": "^1.0.0-beta.20190618-0639",
  +  "@qooxdoo/framework": "^6.0.0-beta-20200609-0430",
  +  "@qooxdoo/compiler": "^2.0.0",
```

The version of compiler is changed and so at next compile I have several message of qx package which doesn't support v2 of the compiler.

hkollmann:
Have you run qx pkg update && qx pkg upgrade?

rommni:
No only qx package migrate like asked in the output of qx compile. Do you want me to test with update and upgrade to see the result ?

rommni:
Fixed by #739

PR #739, opened 2020-07-28
fix issue with version as discussed on gitter

hkollmann:
generate the version info during compiling the compiler

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,13 +13,15 @@ notused/
 node_modules/
 
 # qooxdoo
-/.package-cache.json
-/html/
-/lib/
-/qooxdoo/
-/compiled/
-/qx_packages/
+.package-cache.json
+html/
+lib/
+qooxdoo/
+compiled/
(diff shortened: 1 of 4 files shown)


rommni:
@cboulanger  The bug was that the version of the compiler was read directly in the package.json, however when the compiler was compiled (in build target) the relative path indicate in the version class:  `var pkg = require("../../../../../package.json");` didn't anymore link to the package.json of the compiler but the package.json of the app using qooxdoo. (Because of moving the class definition to index.js)

So if you have a version (different of 1.0.0-beta) in your app's package.json  when you do for example a compile you will get an error asking you to do a package migrate because of the difference of version. When you do the package migrate, the compiler version in the Manifest.json will be change to  the version of your app. And it certainly can cause other problems I didn't discover so far ;)

hkollmann:
fixes https://github.com/qooxdoo/qooxdoo-compiler/issues/737

PR #742, opened 2020-07-31
Add missing babel-eslint

hkollmann:


cboulanger:
Ignoring stalled travis check (which should not be there in the first place), merging.

PR #744, opened 2020-08-03
Fix apiviewer

hkollmann:
- only show compiler api in apiviewer …
- use deploy to deploy apiviewer
- update apiviewer


cboulanger:
thanks @hkollmann !

PR #747, opened 2020-08-03
Fix #746

### LINKED ISSUE #746: load() not returning the config anymore broke my application
#735 stopped load from returning the configuration

oetiker:
load should return the current configuration

### DIFF
--- a/source/class/qx/tool/cli/api/AbstractApi.js
+++ b/source/class/qx/tool/cli/api/AbstractApi.js
@@ -32,15 +32,20 @@ qx.Class.define("qx.tool.cli.api.AbstractApi", {
     rootDir: {
       check: "String",
       nullable: false
-    }
+    },
+    /** Configuration data for the compiler */
+    configuration: {
+    },
   },
   
   members: {
     /**
-     * Loads the configuration data
+     * Loads the configuration
+     *
+     * @return {Map} configuration data
      */
     async load() {
-      // Nothing
+      return this.getConfiguration();
     },
     
     /**
--- a/source/class/qx/tool/cli/api/AbstractApi.js
+++ b/source/class/qx/tool/cli/api/AbstractApi.js
@@ -35,7 +35,7 @@ qx.Class.define("qx.tool.cli.api.AbstractApi", {
     },
     /** Configuration data for the compiler */
     configuration: {
-    },
+    }
   },
   
   members: {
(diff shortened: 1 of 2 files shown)


source/class/qx/tool/cli/api/AbstractApi.js
```diff
@@ -32,15 +32,20 @@ qx.Class.define("qx.tool.cli.api.AbstractApi", {
     rootDir: {
       check: "String",
       nullable: false
-    }
+    },
+    /** Configuration data for the compiler */
+    configuration: {
+    },
```
oetiker, line 38:
```suggestion
    }
```

hkollmann:
Here the log from the test:
Error: Property configuration of an instance of qx.tool.cli.api.AbstractApi is not (yet) ready!
    at wrapper.eval [as getConfiguration] (eval at __P_96_13 (/home/runner/work/qooxdoo-compiler/qooxdoo-compiler/lib/compiler/index.js:13154:36), <anonymous>:3:166)

getConfiguration is called before the class is constructed completely. I remember - that's was the reason for the breaking change.
```
await load();
let config = this,getConfiguration()
```
should work

source/class/qx/tool/cli/api/AbstractApi.js
```diff
@@ -32,15 +32,20 @@ qx.Class.define("qx.tool.cli.api.AbstractApi", {
     rootDir: {
       check: "String",
       nullable: false
+    },
+    /** Configuration data for the compiler */
+    configuration: {
```
oetiker, line 37:
```suggestion
    configuration: {
      init: {}
```

oetiker:
while investigating this one, I found that node_modules contains: `node_modules/@qooxdoo/compiler/node_modules/@qooxdoo/compiler/source/class/qx/tool/cli/api/AbstractApi.js` which it probably should not, AND which is even from an older version of the code ...

oetiker:
> Here the log from the test:
> Error: Property configuration of an instance of qx.tool.cli.api.AbstractApi is not (yet) ready!
> at wrapper.eval [as getConfiguration] (eval at __P_96_13 (/home/runner/work/qooxdoo-compiler/qooxdoo-compiler/lib/compiler/index.js:13154:36), :3:166)
> 
> getConfiguration is called before the class is constructed completely. I remember - that's was the reason for the breaking change.
> 
> ```
> await load();
> let config = this.getConfiguration()
> ```
> 
> should work

well, I guess there is a better way ....

hkollmann:
I tried it half a day but did not find one.
Good luck!

Issue #750, opened 2020-08-04
qx package migrate incorrectly altered "extends" member in "esLintConfig"


rad-pat:
Using compiler @qooxdoo/compiler@1.0.0-beta-20200803-1413, upon compile it told me I needed to run `qx package migrate`. Upon running this command, it changed the `"extends"` member in the `"esLintConfig"` within my compile.json from `"@qooxdoo/qx/browser"` to `"@qooxdoo/@qooxdoo/qx/browser"`, preprending an additonal `@qooxdoo`

hkollmann:
Could you provide a small test project? Or - better - debug the migrate command and provide a PR?

cboulanger:
Problem is here: https://github.com/qooxdoo/qooxdoo-compiler/blob/1925a39f187c1ff74f84082810574e5cd539a29f/source/class/qx/tool/cli/commands/package/Migrate.js#L224-L223
I can provide a fix.

PR #751, opened 2020-08-04
Fix for #750

cboulanger:


cboulanger:
The GIT history is messed up, closing

Issue #760, opened 2020-09-21
webfont loading issue


oetiker:
When using the "provides/webfonts" key in `Manifest.json` to configure webfonts, the resulting application tries to load these fonts, regardless of whether they have actually been mentioned anywhere in the code. Especially it tries to load them despite them not being mentioned in @assets ... 

currently we have the following in the framework/Manifest.json of qooxdoo

```json
  "provides" :
  {
    "namespace"   : "qx",
    "encoding"    : "utf-8",
    "class"       : "source/class",
    "resource"    : "source/resource",
    "translation" : "source/translation",
    "type"        : "library",
    "webfonts":[
      {
        "name": "MaterialIcons",
        "defaultSize": 32,
        "comparisonString": "\ue1fe\uef39",
        "resources": [
          "qx/iconfont/MaterialIcons/materialicons-v50.eot",
          "qx/iconfont/MaterialIcons/materialicons-v50.woff2",
          "qx/iconfont/MaterialIcons/materialicons-v50.woff",
          "qx/iconfont/MaterialIcons/materialicons-v50.ttf"
        ]
      },
      {
        "name": "MaterialIconsOutlined",
        "defaultSize": 32,
        "comparisonString": "\ue1fe\uef39",
        "resources": [
          "qx/iconfont/MaterialIcons/materialiconsoutlined-v18.eot",
          "qx/iconfont/MaterialIcons/materialiconsoutlined-v18.woff2",
          "qx/iconfont/MaterialIcons/materialiconsoutlined-v18.woff",
          "qx/iconfont/MaterialIcons/materialiconsoutlined-v18.otf"
        ]
      },
      {
        "name": "MaterialIconsRound",
        "defaultSize": 32,
        "comparisonString": "\ue1fe\uef39",
        "resources": [
          "qx/iconfont/MaterialIcons/materialiconsround-v18.eot",
          "qx/iconfont/MaterialIcons/materialiconsround-v18.woff2",
          "qx/iconfont/MaterialIcons/materialiconsround-v18.woff",
          "qx/iconfont/MaterialIcons/materialiconsround-v18.otf"
        ]
      },
      {
        "name": "MaterialIconsSharp",
        "comparisonString": "\ue1fe\uef39",
        "defaultSize": 32,
        "resources": [
          "qx/iconfont/MaterialIcons/materialiconssharp-v19.eot",
          "qx/iconfont/MaterialIcons/materialiconssharp-v19.woff2",
[...]

PR #772, opened 2020-11-21
fix for https://github.com/qooxdoo/qooxdoo-compiler/issues/760: webfo…

hkollmann:
…nt loading issue

WebFonts are only loaded if they have an @assert statement somewhere

### DIFF
--- a/source/class/qx/tool/compiler/targets/Target.js
+++ b/source/class/qx/tool/compiler/targets/Target.js
@@ -401,12 +401,20 @@ qx.Class.define("qx.tool.compiler.targets.Target", {
         });
       });
       
+      var assetUris = application.getAssetUris(t, rm, appMeta.getEnvironment()); // Save any changes that getAssets collected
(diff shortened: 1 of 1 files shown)


source/class/qx/tool/compiler/targets/Target.js
```diff
@@ -401,21 +401,40 @@ qx.Class.define("qx.tool.compiler.targets.Target", {
...
       requiredLibs.forEach(libnamespace => {
         var library = analyser.findLibrary(libnamespace);
         var fonts = library.getWebFonts();
         if (!fonts) {
           return;
         }
+
         const loadFont = async font => {
           try {
+            // check if font is assert somewhere
```
cboulanger, line 428: Typo: s/assert/asset/

cboulanger (approved):

oetiker (approved):

cboulanger:
Thank you @hkollmann !!!

PR #778, opened 2020-12-04
fix linter issues

hkollmann:


source/class/qx/tool/cli/api/CompilerApi.js
```diff
@@ -56,7 +56,7 @@ qx.Class.define("qx.tool.cli.api.CompilerApi", {
     
     /**
      * Register compiler tests
-     * @param  cmd: {qx.tool.cli.commands.Command} current command
```
cboulanger, line 59: This is a loss of type information here - @johnspackman , have you consistently formatted jsdoc param type this way - then a regex search/replace could preserve the type info

source/class/qx/tool/cli/api/CompilerApi.js
```diff
@@ -56,7 +56,7 @@ qx.Class.define("qx.tool.cli.api.CompilerApi", {
...
+     * @param {*} cmd current command
      * @return {Promise<void>}
```
cboulanger, line 60: So the linter does not accept  hinting at the type of the resolved Promise value?

source/class/qx/tool/cli/commands/Package.js
```diff
@@ -131,7 +131,7 @@ qx.Class.define("qx.tool.cli.commands.Package", {
 
     /**
      * Convenience method to return all config file models as an array
-     * @return {Promise<[{qx.tool.config.Manifest}, {qx.tool.config.Lockfile}, {qx.tool.config.Compile}]>}
+     * @return {Promise}  [qx.tool.config.Manifest, qx.tool.config.Lockfile]
```
cboulanger, line 134: This also seems to have corrupted the type hint

source/class/qx/tool/compiler/ClassFile.js
```diff
@@ -2335,7 +2402,7 @@ qx.Class.define("qx.tool.compiler.ClassFile", {
     /**
      * Returns the list of required classes (dependencies)
      *
-     * @returns
+     * @return
```
cboulanger, line 2405: Isn't an empty return statement redundant?

cboulanger (approved):

source/class/qx/tool/cli/api/CompilerApi.js
johnspackman: possibly, but what's wrong with the way it was before?  I thought that you were supposed to use a fully dotted classname unless the class being referred to is in the same package as this class

source/class/qx/tool/cli/api/CompilerApi.js
johnspackman: i think that hinting is non standard, it's one of my bad habits and the function should be expressed as `async` and no return value

source/class/qx/tool/cli/commands/Package.js
johnspackman: in es6, the function should be `async`

source/class/qx/tool/cli/Cli.js
```diff
@@ -549,6 +549,9 @@ Versions: @qooxdoo/compiler    v${qx.tool.compiler.Version.VERSION}
      * @return {Boolean}
      */
     isExplicitArg(option) {
+      /**
+       * @param option
+       */
```
johnspackman, line 554: do we have to enable comments for mini helper functions like this?  They're normally pretty self explanatory, like this one

source/class/qx/tool/cli/commands/Command.js
```diff
@@ -128,7 +128,7 @@ qx.Class.define("qx.tool.cli.commands.Command", {
      * Returns the path to the current library. If the current directory contains several libraries,
      * the first one found is returned.
      * @throws {Error} Throws an error if no library can be found.
-     * @return {Promise<String>)} A promise that resolves with the absolute path to the library
+     * @return {Promise} A promise that resolves with the absolute path to the library
```
johnspackman, line 131: This is wrong, it should be just `@return {String}`

johnspackman (changes requested):

source/class/qx/tool/cli/commands/Command.js
cboulanger: Hm, an async function is always returning a Promise. I don't think that the JSDoc `@return` statement in async functions makes this implicit and states the return value only (since the Promise can also be rejected). Since the function is async, it should be documented to return a Promise, don't you think?

johnspackman: AIUI there isnt an official jsdoc for `Promise<String>`, but IMHO the asynchronous nature of the function is what the `async` modifier is for; it is a language-level feature to describe that the function returns a promise, so we do not need to duplicate that in the jsdoc.  From the other point of view, if you have an async function, would it need to have it's `@return` documented as a Promise?  I think that would be a bit unnatural, because the async/await keywords are there to blend Promises into the background

Whether a promise is rejected or an exception can be thrown is not documented by `@return` either, so I don't think that's relevant (i think that there is a `@throws` available but its not used in the framework)

oetiker:
wow

source/class/qx/tool/cli/api/CompilerApi.js
```diff
@@ -56,8 +56,8 @@ qx.Class.define("qx.tool.cli.api.CompilerApi", {
     
     /**
      * Register compiler tests
-     * @param  cmd: {qx.tool.cli.commands.Command} current command
-     * @return {Promise<void>}
+     * @param {qx.tool.cli.commands.Command} cmd  current command
+     * @async
```
johnspackman, line 60: why are we adding `@async` in the jsdoc?  If this is needed for API viewer etc, it should be output as the meta data by the compiler, because the jsdoc will always be out of date and unreliable.

source/class/qx/tool/cli/commands/Command.js
```diff
@@ -334,6 +335,9 @@ qx.Class.define("qx.tool.cli.commands.Command", {
      * @return {Boolean}
      */
     isExplicitArg(option) {
+      /**
+       * @param option
+       */
```
johnspackman, line 340: imho private functions like this do not need comments unless there is a real need for it - and adding a blank comment is not really helpful, it just adds cruft.

source/class/qx/tool/cli/commands/Package.js
```diff
@@ -131,8 +131,9 @@ qx.Class.define("qx.tool.cli.commands.Package", {
 
     /**
      * Convenience method to return all config file models as an array
-     * @return {Promise<[{qx.tool.config.Manifest}, {qx.tool.config.Lockfile}, {qx.tool.config.Compile}]>}
+     * @return {Promise<qx.tool.config.Manifest, qx.tool.config.Lockfile>}
```
johnspackman, line 134: should be `@return [] an array containing {qx.tool.config.Manifest} and {qx.tool.config.Lockfile}`

source/class/qx/tool/cli/commands/Package.js
```diff
@@ -131,8 +131,9 @@ qx.Class.define("qx.tool.cli.commands.Package", {
 
     /**
      * Convenience method to return all config file models as an array
-     * @return {Promise<[{qx.tool.config.Manifest}, {qx.tool.config.Lockfile}, {qx.tool.config.Compile}]>}
+     * @return {Promise<qx.tool.config.Manifest, qx.tool.config.Lockfile>} 
      * @private
+     * @async
```
johnspackman, line 136: not needed etc

johnspackman (commented):
I feel like I'm being overly critical and nit-picking with my comments above, I really don't mean to be obstructive because I do appreciate that the intention is to improve the compiler code (and it would definitely benefit from it, there's a lot of code in there that I wrote that's a bit, um, sloppy 😊 as well as not taking advantage of ES6 constructs like `async` that make such a difference)

At the same time though, I'm not sure what the benefit of a lot of lint's automatic changes are; function comments are meaningless unless filled in, and on private/inner functions the comment would often be stating obvious and therefore pointless (and of course, something to be manually maintained in parallel to the code and which can get out of sync)

Another good example of this is where a function overrides another - the specification of the function parameters and return type are determined by the base class (or interface), and not the derived class.  In Java, the derived-class method would just be tagged with `@Override`, which is an annotation that is used to tell the compiler that we know that there is a base method (ie that we know to call the equivalent of `this.base(arguments)`), and in Qooxdoo code I tend to use a similar non-existant JsDoc `/** @Override */`

Finally, where there is a property with an apply method, and the method is called "_applyPropertyName", do we really have to spell out the purpose of `(value, oldValue, name)` each time?  Is it not enough to say `/** Apply for "propertyName" */`

Can we turn off mandatory comments for inner functions?  Or turn them off completely?

source/class/qx/tool/cli/api/CompilerApi.js
cboulanger: I agree, jsdoc should add what cannot be taken from the code itself. Async functions are defined by returning a Promise, so the attention should be on annotating the return value. It would be great if we could update the API viewer to document the type of the value returned by the Promise one day, because that's what really matters for the API consumer.

johnspackman (changes requested):
I've just done some work in a PR I'm about to submit, and in doing so decided to take a look at some of the lint issues.  I disabled the requirement for JSDoc comments before functions and resolved all of the remaining issues - however, in doing so I discovered a couple of bugs that appear to have been introduced after the linting was disabled, and which linting would have picked up.

The bug was that some code was moved, and the indentation was wrong; but the curly-bracketting was also wrong, and the `qx lint --fix` would have corrected the indentation and assumed that the curly bracket was correct, ie `qx lint --fix` would have hidden the bug, which `qx lint` was reporting.

The one time I've used `qx lint --fix` it broke my code so I'm a bit biased :)  but IMHO `qx lint --fix` should never be used, and every issue should be individually reviewed.

Also, as I have a lint-free version in a separate PR, please can this PR be closed without merging?

hkollmann:
https://github.com/qooxdoo/qooxdoo-compiler/pull/782 is preferred

PR #782, opened 2020-12-17
Inline externals

johnspackman:
* adds support for inlining external scripts into the application's index.js - see `qx compile --inline-external-scripts` and in `compile.json` set the target's `"inline-external-scripts": true`;
* adds support for classes to require an external script with an `@external` jsdoc - the external script is only loaded if the class is required by the application;
* added comments for some classes (lint fix);
* restored lost unit tests
* qx lint now passes

This will replace https://github.com/qooxdoo/qooxdoo-compiler/pull/778


cboulanger (approved):
Looks good to me

oetiker (approved):
love it!

package.json
```diff
@@ -8,10 +8,16 @@
   "scripts": {
     "devtools": "node source/resource/qx/tool/bin/build-devtools",
     "website": "node source/resource/qx/tool/bin/build-website",
-    "pretest": "npx qx deploy --target=build --clean",
-    "test": "node bin/qx test --target=build --output-path-prefix=tmp",
+    "test": "cross-os test.os",
```
hkollmann, line 11: The idea here was to use the built in qx test framework instead off a handwritten shell script

johnspackman: the problem is that a whole load of unit tests had been dropped, as well as linting, and it also meant that we had eliminated the step where we compile the compiler twice, once with the `node_modules` compiler and then with the output of the first compile; that's important because it's an additional validation to make sure that we dont release a broken compiler.

The tests which test the compiled version of the compiler were not using the compiled version (ie they were not using the one which had been output into `tmp`)

cboulanger: @hkollmann But isn't the problem that the compiler-based tests aren't fully equivalent to the shell-script-based one? I.e. the shell scripts have to be converted to NodeJS first before we can switch to the "native" testing?

cboulanger: Or at least the shell scripts have to be called from Node - or were they?

johnspackman: there is two things, one is that the tests had not been converted, and the other is that there is a series of steps which need to be performed before starting the tests.  That series of steps could be implemented as a .js file so that we're not using platform dependent shell scripts, but that can be done later - these fixes are restoring the lost tests and test functionality

hkollmann: > The tests which test the compiled version of the compiler were not using the compiled version (ie they were not using the one which had been output into `tmp`)

That's not true - the tests should use the qx command in the test folder. Then thy test the vesion in tmp.

hkollmann: A lot of tests are done with qx test - all scripts in test/compiler are perfomed by the compile.js. 
Only missing tests are in _todo_move_to_unit_test. When i moved to qx test i had no idea how to handle this stuff. Some of them are real unit tests. Some of them should be a test script in the compiler dir.

My fear is if return to batch based testing the work to move to qx test will never be done again.

cboulanger: I think it would be important to establish the `qx test` way and eat our own dog food in that respect as well. What about we reserve say half a day sometime and finish this up in one go, splitting work (like converting the todo stuff? Then we can help each other out in real time when problems occur and we get it over with finally?

johnspackman: Looking at the code, it the compile.js runs the equivalent of `cd test/compiler ; node test-dependencies.js` - and `test-dependencies.js` uses `require("../utils")`; but in https://github.com/qooxdoo/qooxdoo-compiler/blob/master/test/utils.js#L16 it looks for `qx` in the current directory, which does not exist [in the repo](https://github.com/qooxdoo/qooxdoo-compiler/tree/master/test) - this probably explains some of my issues, because I have a `qx` in that directory which uses the version in `bin`.

The way it works now is that the tests which you wrote and integrated into `compile.js` still work, and work as before; what I didn't realise until last night was the PR which introduced them also removed a number of other tests etc.

`npm test` now includes all of the tests, as well as the new ones - I appreciate that it is not cross platform, but reverting that part of the previous PR seems like the fastest way to get them back.  If necessary, the shell script can be re-written as a node.js script?

As an aside, we now have 4 different compiled versions of the compiler which are being used at the same time - `compiled/source`, `compiled/build`, `lib` and `tmp`.  The script now only uses tmp as a temporary staging process before re-compiling, and means that the unit tests are run using the same compiled as the `bin/qx` command (which after all is the command you're likely to use when developing/testing, and if not it would be `bin-src/qx`).

Thinking about, it would be cleaner to get rid of `lib` also and only have that in use during actual deployment (which AFAICR is handled elsewhere now); this would eliminate an extra compile step during development and mean that there are only two versions, `compiled/source` and `compiled/build`, which is more along the lines of what would be expected.  

As I mentioned, I think that some of the issues I had to track down were caused by different compiled versions being used at different times, which is quite time consuming and distracting, so I think it would be an advantage to make thsi simplification.

hkollmann: test/qx is generated in compile.js before testing:
```
        let compilerPath = path.resolve(path.join(maker.getOutputDir(), "compiler"));
        let cmd ="#!/usr/bin/env node\n" + `require("${compilerPath}");\n`;
        fs.writeFileSync("test/qx", cmd, {mode: 0o777});
```
Could be removed by an static file.

My  workflow intension for testing once was this:

1) bootstrap: deploy qx to /lib. 
    This is needed for the next step because /bin/qx is searching for index.js in the deploy dir. The lib dir 
    is the result of deploying - not compiling.
2) compile a version with the fresh build compiler to the /tmp dir.
3)  run  the tests against the version in the self compiled tmp dir.

johnspackman: > test/qx is generated in compile.js before testing:

Aha, that explains why my qx has an absolute path to my folder - I assumed that it was something I'd hacked in at some point.  My one has `require("/Users/john/dev/Qooxdoo/QxCompiler/compiled/node/build/compiler");` which is not the deployed version.

IMHO it should not be possible to end up with differently compiled versions of the compiler on the same system; there's no need for a deployment stage during testing, and if I run `qx` from the command line it should do exactly the same as when the scripts run it.  I've literally spent hours just trying to get a unit test in place and running consistently, starting with discovering that the majority of my tests had been removed.  EG the first PR test run crashed when it was running on my machine (there were a load of bugs in the release too, but that's another story! 😊 )

How about this then - 
* reinstate the `pretest` rule and make it double/dogfood compile the compiler
* remove `lib` (although leave it as a deployment directory) and provide two versions of `qx`, either of which can be added to someone's PATH, one is for `compiled/source` and the other is for `compiled/build`
* check that the deployment repo will still have a valid `qx` that matches the `lib` directory created during deployment
* invert the unit tests so that the recovered tests are triggered by `qx test` rather than the other way around

This would eliminate all OS-dependent scripts, and mean that running `qx` on the command line is exactly the same compiler as used by all tests

hkollmann: +1

cboulanger (approved):

hkollmann (approved):

cboulanger:
@johnspackman Great feature addition! Could you add the new inlining feature to https://github.com/qooxdoo/documentation/blob/master/development/howto/using_non_qx_libs.md ? This section is more accessible to newcomers than the compiler API documentation. Thank you!

johnspackman:
> @johnspackman Great feature addition! Could you add the new inlining feature to https://github.com/qooxdoo/documentation/blob/master/development/howto/using_non_qx_libs.md ? 

done, and in compiler/configuration

@/all this now passes the unit tests and is ready for merge

PR #783, opened 2020-12-20
fixes for windows

hkollmann:
- enable npm test in window systems
- enable npm link after run on bootstrap-compiler
- add some doc how to build the github version


cboulanger:
Should the `.vscode` file really be in there? I think that we shouldn't commit IDE configuration files to the repo since every developer has their own configuration.

bootstrap-compiler
```diff
@@ -6,7 +6,7 @@ const fs = require("fs");
 (async function() {
   console.log("pwd=" + process.cwd());
   let files = await fs.promises.readdir(".");
-  console.log("files=" + files.join("\n  "));
+  console.log("files=\n  " + files.join("\n  "));
```
johnspackman: all of this could be removed, it was only used for debugging a workflow problem and I should have removed it before

bin/tools/utils.js
```diff
@@ -224,6 +216,5 @@ module.exports = {
   bootstrapCompiler,
   deleteRecursive,
   safeDelete,
-  fsPromises,
-  promisify
```
johnspackman, line 228: whats wrong with this?

bin/tools/utils.js
```diff
@@ -203,13 +195,13 @@ require(path.join(__dirname, "compiled", "node", "build", "compiler"));
...
   console.log("Compiling source version");
-  result = await runCommand(".", "./tmp/qx", "compile");
+  result = await runCommand(".", "node", "./tmp/qx", "compile", "--clean", "-M");
   if (result.exitCode) {
     process.exit(result.exitCode);
   }
   
   console.log("Compiling build version");
-  result = await runCommand(".", "./tmp/qx", "compile", "--target=build");
+  result = await runCommand(".", "node", "./tmp/qx", "deploy", "--target=build", "--clean", "-M");
```
johnspackman, line 204: I don't agree with this; during development there is no need for three different compiled versions of the compiler, and making it so that a `qx deploy` step has to be completed in order to do code/test/debug slows down the whole process.  The deloyment version is *only* required when creating a deployment to npm, ie it is only needed by the https://github.com/qooxdoo/deployment repo.

The way it is set up now, anyone can add either `bin/source` (or `bin/build`) onto their PATH and then they have the version of the compiler under development.  When working on the compiler itself, you `cd` to the compile directory and run `qx compile --watch` (possibly with `--target=build` if you want to test the build version), and that is the fastest way to get it running.

While there are some occasions where the `build` version requires a separate test/development cycle to `source`, 99% of the time the `source` version is by far the most preferable.  IMHO it is really time consuming and distracting to have lots of different versions knocking around because the possibility for the wrong (and out of date) compiler version to be run is high, meaning that you spend time debugging problems only to find that your new code isn't working because it's in the other, or the other, other compiler.

Thinking about it, I now think that possibly the `build` version should not be compiled *at all* except when running `npm test`.

This should mean that you can run any test or other code just by using `qx` - you do not have to remember to use any specific version, and if there is a script (such as a unit test) which runs `qx`, then it will also use the current version under development.

johnspackman: also, same question about `-M`

cboulanger (approved):

johnspackman (changes requested):

bin/tools/utils.js
hkollmann: ```qx test``` is testing the source version. To test the build version you need ```qx test --target=build```.
```qx test``` will do the compile step if it's necessary - so we can change "test": "node ./tmp/qx --target=build" to do the test.

In this case the bootstrap process only has to compile the source version after the bootstrap compiler

Comment (johnspackman) [reply bin/tools/utils.js:204]: My point is that the code has been changed to:
```
result = await runCommand(".", "node", "./tmp/qx", "deploy", "--target=build", "--clean", "-M");
```
This should be `compile` *not* `deploy`

File: bin/tools/utils.js [hunk 4]
```diff
@@ -203,17 +195,17 @@ require(path.join(__dirname, "compiled", "node", "build", "compiler"));
    *  it does not matter if they use source or build, just make sure it is up to date
    */
   console.log("Compiling source version");
-  result = await runCommand(".", "./tmp/qx", "compile");
+  result = await runCommand(".", "node", "./tmp/qx", "compile", "--clean", "-M");
```
hkollmann, line 198: @johnspackman : Would'nt it be enough to build the compiler?

bin/tools/utils.js
```diff
@@ -203,17 +195,17 @@ require(path.join(__dirname, "compiled", "node", "build", "compiler"));
...
   if (result.exitCode) {
     process.exit(result.exitCode);
   }
-  
+
   console.log("Compiling build version");
-  result = await runCommand(".", "./tmp/qx", "compile", "--target=build");
+  result = await runCommand(".", "node", "./tmp/qx", "compile", "--target=build", "--clean", "-M");
```
hkollmann, line 204: Same here?

hkollmann: For testing --app-name=compiler,compilerLibrary is needed

bin/tools/utils.js
johnspackman: I think we only need the compiler at this stage - the testing happens using the re-compiled compiler, rather than the tmp compiler.  Please can you remove the `-M`?

johnspackman (approved):

johnspackman:
lets get this merged and the deployment running again, and sort out any details later?

PR #788, opened 2021-01-13
enforce minimum api viewer version

johnspackman:


cboulanger:
this is probably necessary to save us numerous bug reports by users, but seems a bit ad-hoc-ish. Unfortunately, we don't have a way of enforcing the version of optional dependencies...

source/class/qx/tool/cli/Cli.js
```diff
@@ -387,6 +387,17 @@ Versions: @qooxdoo/compiler    v${qx.tool.compiler.Version.VERSION}
       if (lockfileContent.libraries) {
         config.packages = {};
         lockfileContent.libraries.forEach(function(library) {
+          if (library.uri == "qooxdoo/qxl.apiviewer") {
+            let m = library.repo_tag.match(/^v([0-9]+)\.([0-9]+)\.([0-9]+)$/);
+            if (m) {
+              m.shift();
+              m = m.map(v => parseInt(v, 10));
+              if (m[0] <= 1 && m[1] == 0 && m[2] < 15) {
+                qx.tool.compiler.Console.log("API Viewer is out of date and must be upgraded - please run 'qx package upgrade'");
```
hkollmann, line 396: I suggest:
 please run 'qx package update; qx package upgrade'

hkollmann (commented):
compiler needs apiviewer update too :smile: !

cboulanger:
Maybe a `optionalDependencies` key in Manifest.json would be in order. But the problem here is that we are talking about a dependency of the **compiler** on an optional dependency of the **application**.

cboulanger (approved):
I approve, but we should create an issue to address this class of problems in a more generalizable way. Having to add ad-hoc checks with ad-hoc versions doesn't seem a good solution.

johnspackman:
> this is probably necessary to save us numerous bug reports by users, but seems a bit ad-hoc-ish. Unfortunately, we don't have a way of enforcing the version of optional dependencies...

I completely agree - IMHO there should be a way to invalidate previous package versions, eg a package should be able to declare something along the lines of "don't use a version before x.y.z" and have the compiler emit a warning when this is discovered.

However, implementing that would be a much bigger task and it would be much easier to add that (and test it) if the packaging code was refactored to an API first; at the moment, the packaging is built into being used as a command, which means that it's quite tricky to interrogate the state of the local database vs the remote, etc; that's also why the code gives a warning, rather than just do the upgrade

> Maybe a optionalDependencies key in Manifest.json would be in order. But the problem here is that we are talking about a dependency of the compiler on an optional dependency of the application.

I don't think it's optional - in this case, the app has an explicit requirement for a package, and that version of the package happens to have a bug in it's compile.js.

We would really benefit from hardening the compiler API to at least track what is going on; to do that, the framework could have timeout detection for `fireEventAsync` and `fireDataEventAsync`, so that if they do not complete within a set time, the framework will output a warning

source/class/qx/tool/cli/Cli.js
johnspackman: done

hkollmann (approved):

PR #790, opened 2021-01-16
adds `preserveEnvironment` setting

johnspackman:


cboulanger (approved):

hkollmann (commented):
What's the problem wit the tests?

johnspackman:
@hkollmann tests were failing because of a bug in bootstrap - that's fixed now, and once this PR is merged the other PR can be updated and then it should pass too (famous last words...)

hkollmann (approved):

PR #796, opened 2021-01-24
get rid of DEBUG: confirmed= message

hkollmann:


cboulanger (approved):

source/class/qx/tool/cli/Watch.js
```diff
@@ -131,7 +131,9 @@ qx.Class.define("qx.tool.cli.Watch", {
           });
         });
       }))).then(() => {
-        qx.tool.compiler.Console.debug(`DEBUG: confirmed=${JSON.stringify(confirmed)}`);
+        if (this.isDebug()) {
```
johnspackman, line 134: do we need this?  ie, shouldn't that be in the `qx.tool.compiler.Console.debug` code?

johnspackman (approved):

source/class/qx/tool/cli/Watch.js
hkollmann: It's your own code. See line 109.

johnspackman: i meant shouldn't the `if (this.isDebug())` be in the `Console.debug` function?  Otherwise it would have to be applied to every call to Console.debug...

hkollmann:
I just copied your Code. The debug flag is only valid in watch

Von meinem iPhone gesendet

PR #797, opened 2021-01-24
Fix test

hkollmann:
- fix failed-fast
- fix require. node packages should be always intalled in the cwd. To allow this you need a node-modules dir
- do pull after push. so the current checkout has actual state


johnspackman (approved):

cboulanger (approved):

cboulanger:
Thank you @hkollmann !

PR #804, opened 2021-01-30
small fix for copyFile

hkollmann:
resolve with from, to like copy. So a valid result is returned if single files are synced


johnspackman (approved):

cboulanger (approved):

cboulanger:
Thank you @hkollmann

PR #805, opened 2021-02-08
Improve warning message on mismatch between framework/compiler versions

cboulanger:
The previous message did not reveal what the mismatch was and gave the wrong recommendation if the installed framework/compiler version was **lower** than the dependencies stated in the Manifest.


johnspackman (approved):

hkollmann (approved):

hkollmann:
Thanks @cboulanger !

PR #808, opened 2021-03-26
cache and use-eslintrc are booleans

voger:
The options `--cache` and `--use-eslintrc` are recognized as `string` and they remain always true. This PR sets them as `boolean` so the yargs translates them properly..


cboulanger (approved):

johnspackman:
Hi,

Thank you for your email - I am unavailable, returning on April 12th.  
 I will get back to you as soon as possible, but if you have an emergency support issue please describe your issue in detail in email and then send a text message to the mobile number you have for me.

Regards
John Spackman

hkollmann:
The compiler is merged with the qooxdoo main project. Could you do the same in qooxdoo? Thanks!

hkollmann (approved):

voger:
@hkollmann Sure. Does this mean that all the compiler pull requests should be directed to the qooxdoo main package?

hkollmann:
Yes.

cboulanger (approved):

PR #836, opened 2022-02-12
fix mixin, see https://github.com/qooxdoo/qooxdoo/issues/10355

hkollmann:


cboulanger (approved):

derrell (approved):
I'll take your word for it

